PR #11, opened 2020-08-14
support h264 with no access unit delimiters

ccbrown:
The 8K stream didn't have access unit delimiters, which is fine, but it requires a bit of extra work to support. We have frame counting logic in a few places, so I've added structures to the h264 and h265 crates for identifying frames. Short 8K clip added as a unit test.


ccbrown:
Updated with optimizations and a benchmark. The first commit here too 97ms for each iteration of the benchmark. The second commit refactors the h264/h265 bitstreams to operate on iterators instead of `AsRef<[u8]>`, which allows us to remove emulation prevention bytes from the RBSP and decode the slice header much more efficiently. After the second commit, the benchmark only takes 5ms per iteration.

PR #14, opened 2020-08-20
improve h265 frame counting

ccbrown:
Basically the same thing as https://github.com/sportsball-ai/av-rs/pull/11 but for H265. Fortunately this is actually much easier for H265 than for H264.

### DIFF
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -557,4 +557,37 @@ mod test {
             ]
         );
     }
+
+    #[tokio::test]
+    async fn test_analyzer_h265_8k() {
+        let mut analyzer = Analyzer::new();
+
+        {
+            let mut f = File::open("src/testdata/h265-8k.ts").unwrap();
+            let mut buf = Vec::new();
+            f.read_to_end(&mut buf).unwrap();
+            let packets = ts::decode_packets(&buf).unwrap();
+            analyzer.handle_packets(&packets).unwrap();
+            analyzer.flush().unwrap();
+        }
+
+        assert_eq!(
+            analyzer.streams(),
+            vec![
+                StreamInfo::Video {
+                    width: 7680,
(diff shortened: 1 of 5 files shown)


ccbrown:
Made the test file larger so that it includes something weird in that stream: a NALU of type 48, which is unspecified by the H.265 spec. I don't know what's in the NALU, but for those unspecified NALUs, I'm now skipping the forbidden zero bit check. It's unclear if this stream violates the H265 spec by setting the forbidden zero bit to one for an unspecified NALU type, but it seems best to allow it even if it is a violation.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -557,4 +557,37 @@ mod test {
...
+            analyzer.flush().unwrap();
+        }
+
+        assert_eq!(
+            analyzer.streams(),
+            vec![
+                StreamInfo::Video {
+                    width: 7680,
+                    height: 4320,
+                    frame_rate: 59.94,
```
sxing, line 580: This should be 29.97? I'm hoping the stream doesn't report at 59.94 though.

sxing:
Added a comment for the framerate of the test file. Should be 29.97? Also, ffmpeg is super pissed about that NALU as well, with the exact same error getting pushed up. I wouldn't be surprised if this was a spec violation. I suspect the manufacturer doesn't have a lot of people out there pushing live 8K H265 on what are essentially cinema cameras.

mpegts-segmenter/src/analyzer.rs
ccbrown: The parameter sets indicate 59.94. FFmpeg also thinks this file is 59.94 FPS. 🤷‍♂️

ccbrown:
@sxing I think the actual framerate is 29.97, even though the parameter sets indicate 59.97. I'm not really sure if there's anything better that we can do about that.

PR #26, opened 2020-10-18
iterate over u8 instead of &u8

ccbrown:
This makes all of the H26X parsing use an iterator over `u8` instead of `&u8`. The rationale being:

1. It's more concise and eliminates lifetime parameters.
2. It facilitates using iterators that own data (which you can't do with `&u8` until GATs become real). So you can just `Bitstream::new(vec![])` for example, where previously you would have had to assign the vector to a temporary variable. Similarly this facilitates iterators that _generate_ data (which we don't have at the moment, but they might be useful for situations such as fuzz testing).

Because any iterator over `&u8` can be trivially turned into an iterator over `u8` with `.copied()`, there's no loss in functionality.

### DIFF
--- a/h264/src/bitstream.rs
+++ b/h264/src/bitstream.rs
@@ -6,8 +6,8 @@ pub struct Bitstream<T> {
     next_bits_length: usize,
 }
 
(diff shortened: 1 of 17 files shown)


dotcypress (approved):
👍

PR #28, opened 2020-10-19
Timecodes

dotcypress:
This PR contains `Timecode` extraction support for `h264`.


h264/src/sequence_parameter_set.rs
```diff
@@ -311,6 +311,13 @@ pub struct VUIParameters {
     pub time_scale: U32,
     pub fixed_frame_rate_flag: U1,
     // }
+    pub nal_hrd_parameters_present_flag: U1,
+    pub vcl_hrd_parameters_present_flag: U1,
+    pub low_delay_hrd_flag: U1,
+    pub pic_struct_present_flag: U1,
+
+    pub nal_hrd_parameters: Option<HRDParameters>,
+    pub vcl_hrd_parameters: Option<HRDParameters>,
```
ccbrown, line 320: For all of the other structs and fields, I've been keeping them in order and commented with their conditionals, making them read as closely as possible as they do in the spec. See the other fields in the same struct for example. These look like they're out of order and missing comments.

utils/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
```
ccbrown, line 1: This package looks unintentional.

h264/src/lib.rs
```diff
@@ -115,6 +118,35 @@ pub fn read_annex_b<T: Read>(reader: T) -> ReadAnnexB<T> {
     }
 }
 
+pub fn read_first_sei_pic_timing<T: Read>(reader: T) -> io::Result<PicTiming> {
```
ccbrown, line 121: This is extremely niche. It only works for annex B, only works when you don't care about anything else in the bitstream, only works when you don't care about subsequent timecodes, only works when there's an SPS NALU present before the first SEI, etc. And reading through the entire stream again just for that one bit of information is far from zero-cost.

I'm thinking that this is better done elsewhere. For example, the analyzer in mpegts-segmenter can do this more efficiently and output it in its `StreamInfo` for the video.

ccbrown (changes requested):

h264/src/sei.rs
```diff
@@ -0,0 +1,129 @@
...
+
+pub const SEI_PAYLOAD_TYPE_PIC_TIMING: u64 = 1;
+
+// ITU-T H.264, 04/2017, 7.3.2.3.1
+#[derive(Clone, Debug, Default)]
+pub struct SEI {
+  pub payload_type: u64,
+  pub payload_size: u64,
+  pub pic_timing: Option<PicTiming>,
+}
```
ccbrown, line 13: I've been trying to keep the naming consistent with the spec. Can we name this `SEIMessage`?

h264/src/sei.rs
```diff
@@ -0,0 +1,129 @@
...
+
+#[derive(Clone, Debug, Default)]
+pub struct PicTiming {
+  pub hours: u8,
+  pub minutes: u8,
+  pub seconds: u8,
```
ccbrown, line 19: How does a user of this library know if hours, minutes, or seconds are even present? Can we add the fields for sub-second timecodes like the frame count?

mpegts-segmenter/src/analyzer.rs
```diff
@@ -75,6 +77,11 @@ impl Stream {
                     access_unit_counter.count()
                 },
                 rfc6381_codec: rfc6381_codec.clone(),
+                timing: pic_timing.as_ref().map(|timing| StreamTiming {
```
ccbrown, line 80: Let's name this field "timecode" instead of "timing" to be more specific. I'll be adding other timing-related fields soon.

ccbrown (changes requested):
I think there's quite a bit missing here. We need to support the full timecode, including frame number. Only decoding the seconds isn't enough. Furthermore, we need to properly support pic_timing messages that don't contain hours / minutes / seconds.

h264/src/sei.rs
dotcypress: Yep, will do

h264/src/sei.rs
```diff
@@ -0,0 +1,132 @@
...
+  pub payload_size: u64,
+  pub pic_timing: Option<PicTiming>,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct PicTiming {
+  pub hours: Option<u8>,
+  pub minutes: Option<u8>,
+  pub seconds: Option<u8>,
+  pub frames: Option<u8>,
```
ccbrown, line 20: Is there a reason we shouldn't have the other fields here? The rest of this library aims to be an unopinionated transliteration of the spec, but this `PicTiming` struct is something that differs radically from what's in the spec. In addition to omitting a lot of relevant fields, it doesn't have well-defined behavior when `NumClockTS` isn't 1 (all of these things should be in a `Vec`).

mpegts-segmenter/src/analyzer.rs
```diff
@@ -283,10 +308,19 @@ pub enum StreamInfo {
...
     },
     Other,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct StreamTimecode {
+    pub hours: Option<u8>,
+    pub minutes: Option<u8>,
+    pub seconds: Option<u8>,
+    pub frames: Option<u8>,
```
ccbrown, line 321: Why are all of these `Option`s? We should either have the timecode or not have the timecode. That means you need to implement logic for...

> When hours_value is not present, the previous hours_value in decoding order shall be used as hH to compute clockTimestamp.

> When minutes_value is not present, the previous minutes_value in decoding order shall be used as mM to compute clockTimestamp.

> When seconds_value is not present, the previous seconds_value in decoding order shall be used as sS to compute clockTimestamp.

ccbrown (changes requested):

mpegts-segmenter/src/analyzer.rs
dotcypress: Fixed

mpegts-segmenter/src/analyzer.rs
```diff
@@ -283,10 +316,19 @@ pub enum StreamInfo {
         frame_rate: f64,
         frame_count: u64,
         rfc6381_codec: Option<String>,
+        timecodes: Vec<StreamTimecode>,
```
ccbrown, line 319: I think this could use a comment explaining why it's a `Vec` and what it means for it to contain multiple timecodes.

h264/src/sei.rs
ccbrown: Looks like the first question here got unresolved:

> How does a user of this library know if hours, minutes, or seconds are even present?

If the first pic_timing message we see is missing hours, we don't have a complete timecode. How does a user of this library distinguish missing hours from 0 hours?

Again, the structs in this library should be as unopinionated as possible. We shouldn't make any assumptions about what fields the user of this library will need or what they're doing with them.

Rather than parsing everything and discarding the fields we don't think users will need, I'd prefer if we literally just copied / pasted the spec here. I.e. this struct should have fields named `ct_type`, `nuit_field_based_flag`, `counting_type`, `full_timestamp_flag`, `discontinuity_flag`, `cnt_dropped_flag`, `n_frames`, `seconds_value`, `minutes_value`, etc.

ccbrown (changes requested):

h264/src/sei.rs
```diff
@@ -0,0 +1,146 @@
...
+            }
+          }
+        }
+      }
+
+      if let Some(hrd_params) = hrd_params {
+        if hrd_params.time_offset_length.0 > 0 {
+          timecode.time_offset.0 = bs.read_bits(hrd_params.time_offset_length.0 as usize)? as u16;
+        }
+      }
```
ccbrown, line 139: > When the time_offset_length syntax element is not present, it shall be inferred to be equal to 24.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -283,10 +314,20 @@ pub enum StreamInfo {
         frame_rate: f64,
         frame_count: u64,
         rfc6381_codec: Option<String>,
+        // One timecode per each ClockTS
```
ccbrown, line 317: I meant at a slightly higher level. Let's not make users of this struct dig through the H264 spec to figure out when there might be more than one clock timestamp and what each one means.

ccbrown (changes requested):

mpegts-segmenter/src/analyzer.rs
```diff
@@ -75,6 +77,18 @@ impl Stream {
...
+                timecodes: timecode
+                    .as_ref()
+                    .map(|t| t.timecodes.as_ref())
+                    .unwrap_or(&vec![])
+                    .iter()
+                    .map(|t| StreamTimecode {
+                        hours: t.hours.0,
+                        minutes: t.minutes.0,
+                        seconds: t.seconds.0,
+                        frames: t.n_frames.0,
```
ccbrown, line 89: Where did the logic go for the fallbacks like:

> When minutes_value is not present, the previous minutes_value in decoding order shall be used as mM to compute clockTimestamp.

?

mpegts-segmenter/src/segmenter.rs
```diff
@@ -253,13 +253,15 @@ impl SegmentInfo {
...
                                 rfc6381_codec,
+                                timecodes,
                             },
                         ) => Some(StreamInfo::Video {
                             width: *width,
                             height: *height,
                             frame_rate: *frame_rate,
                             frame_count: if frame_count >= prev_frame_count { frame_count - prev_frame_count } else { 0 },
                             rfc6381_codec: rfc6381_codec.clone(),
+                            timecodes: timecodes.clone(),
```
ccbrown, line 264: It looks like you're using timecodes from the end of the segment here? I think it would be a lot more useful if we had timecodes from the beginning of the segment. E.g. a 3-second segment that starts at 00:05:00.00 should have a timecode of 00:05:00.00 rather than 00:08:00.00.

dotcypress: Now all timecodes from are available

mpegts-segmenter/src/analyzer.rs
```diff
@@ -176,6 +206,19 @@ impl Stream {
...
+                        }
+                        h264::NAL_UNIT_TYPE_SUPPLEMENTAL_ENHANCEMENT_INFORMATION => {
+                            let bs = h264::Bitstream::new(nalu.iter().copied());
+                            let mut nalu = h264::NALUnit::decode(bs)?;
+
+                            if let Some(vui_params) = &last_vui_parameters {
+                                let mut rbsp = h264::Bitstream::new(&mut nalu.rbsp_byte);
+                                let sei = h264::SEIMessage::decode(&mut rbsp, &vui_params)?;
+                                if let Some(pic_timing) = sei.pic_timing {
+                                    timecodes.push(pic_timing);
```
ccbrown, line 219: If every frame in a 5 hour 29.97 FPS stream has a timecode, there will be 539,460 elements in `timecodes` by the end. Keeping every one of those in memory seems a little excessive.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -75,6 +77,30 @@ impl Stream {
                     access_unit_counter.count()
                 },
                 rfc6381_codec: rfc6381_codec.clone(),
+                timecodes: timecodes.iter().filter_map(|t| t.timecodes.iter().next()).fold(vec![], |mut acc, t| {
```
ccbrown, line 80: What's happening here? Why are we just taking the first timecode from each `PicTiming` message?

mpegts-segmenter/src/analyzer.rs
```diff
@@ -587,6 +644,38 @@ mod test {
...
+                            seconds: 29,
+                            frames: 2
+                        },
+                        StreamTimecode {
+                            hours: 18,
+                            minutes: 57,
+                            seconds: 30,
+                            frames: 2
+                        }
+                    ],
```
ccbrown, line 678: Is this actually the expected output? At a glance, it looks like there's an SEI message after every access unit delimiter. Why is there only one timecode per second?

ccbrown (changes requested):
Looks great! Just needs tests!

ccbrown (approved):

PR #31, opened 2020-11-02
Extract timecodes

dotcypress:
Extract timecodes from the video stream.

### DIFF
--- a/h264/src/sei.rs
+++ b/h264/src/sei.rs
@@ -90,7 +90,7 @@ impl PicTiming {
     pub fn decode<T: Iterator<Item = u8>>(bs: &mut Bitstream<T>, vui_params: &VUIParameters) -> io::Result<Self> {
         let mut ret = Self::default();
 
-        let hrd_params = vui_params.nal_hrd_parameters.as_ref().or(vui_params.vcl_hrd_parameters.as_ref());
+        let hrd_params = vui_params.nal_hrd_parameters.as_ref().or_else(||vui_params.vcl_hrd_parameters.as_ref());
         if let Some(hrd_params) = hrd_params {
             ret.cpb_removal_delay = bs.read_bits(hrd_params.cpb_removal_delay_length_minus1.0 as usize + 1)?;
             ret.dpb_output_delay = bs.read_bits(hrd_params.dpb_output_delay_length_minus1.0 as usize + 1)?;
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -218,7 +218,7 @@ impl Stream {
                                             seconds: t.seconds.0,
                                             frames: t.n_frames.0,
                                         };
-                                        if let Some(previous_timecode) = acc.iter().last() {
+                                        if let Some(previous_timecode) = acc.last() {
                                             if t.full_timestamp_flag.0 == 0 {
                                                 if t.seconds_flag.0 == 0 {
                                                     timecode.seconds = previous_timecode.seconds;
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -5,7 +5,7 @@ use std::error::Error;
 pub type Result<T> = std::result::Result<T, Box<dyn Error + Send + Sync>>;
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct StreamTimecode {
+pub struct Timecode {
     pub hours: u8,
     pub minutes: u8,
     pub seconds: u8,
@@ -31,8 +31,8 @@ pub enum Stream {
         is_interlaced: bool,
         access_unit_counter: h264::AccessUnitCounter,
         timecode_pending: bool,
-        timecode: Option<StreamTimecode>,
-        last_timecode: Option<StreamTimecode>,
+        timecode: Option<Timecode>,
+        last_timecode: Option<Timecode>,
         last_vui_parameters: Option<h264::VUIParameters>,
     },
     HEVCVideo {
@@ -205,7 +205,7 @@ impl Stream {
                                 let mut rbsp = h264::Bitstream::new(&mut nalu.rbsp_byte);
                                 let sei = h264::SEI::decode(&mut rbsp)?;
 
-                                let acc: Vec<StreamTimecode> = last_timecode.iter().cloned().collect();
+                                let acc: Vec<Timecode> = last_timecode.iter().cloned().collect();
                                 let timecodes = sei
                                     .sei_message
                                     .iter()
@@ -214,7 +214,7 @@ impl Stream {
                                         h264::PicTiming::decode(&mut bs, &vui_params).unwrap().timecodes
                                     })
                                     .fold(acc, |mut acc, t| {
-                                        let mut timecode = StreamTimecode {
+                                        let mut timecode = Timecode {
                                             hours: t.hours.0,
                                             minutes: t.minutes.0,
                                             seconds: t.seconds.0,
@@ -362,7 +362,7 @@ pub enum StreamInfo {
         frame_rate: f64,
         frame_count: u64,
         rfc6381_codec: Option<String>,
-        timecode: Option<StreamTimecode>,
+        timecode: Option<Timecode>,
     },
     Other,
 }
@@ -683,7 +683,7 @@ mod test {
                     frame_rate: 29.97,
                     frame_count: 152,
                     rfc6381_codec: Some("avc1.4d4028".to_string()),
-                    timecode: Some(StreamTimecode {
+                    timecode: Some(Timecode {
                         hours: 18,
                         minutes: 57,
                         seconds: 26,
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -31,6 +31,7 @@ pub enum Stream {
         is_interlaced: bool,
         access_unit_counter: h264::AccessUnitCounter,
         timecode: Option<StreamTimecode>,
+        timecode_pending: bool
     },
     HEVCVideo {
         pes: pes::Stream,
@@ -160,6 +161,7 @@ impl Stream {
                 is_interlaced,
                 access_unit_counter,
                 timecode,
+                timecode_pending,
                 ..
             } => {
                 use h264::Decode;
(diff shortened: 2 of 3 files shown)


ccbrown (changes requested):
I don't think this is quite right. Let's do some TDD: Can you add a test for the segmenter to take a file with timecodes, segment it, then ensure that each segment has the expected timecode?

mpegts-segmenter/src/segmenter.rs
```diff
@@ -459,10 +461,43 @@ mod test {
...
+                    minutes: 57,
+                    seconds: 28,
+                    frames: 2
+                },
+                analyzer::StreamTimecode {
+                    hours: 18,
+                    minutes: 57,
+                    seconds: 30,
+                    frames: 2
+                }
```
ccbrown, line 499: Please double check these. We want the timecode of the first frame in each segment when possible.

ccbrown (changes requested):

mpegts-segmenter/src/segmenter.rs
```diff
@@ -157,6 +157,8 @@ impl<S: SegmentStorage> Segmenter<S> {
                     pts: None,
                     bytes_written: 0,
                 });
+
+                self.analyzer.reset_timecodes();
```
ccbrown, line 161: I feel like we should be able to do this without this `reset_timecodes` mechanism. At this point, the segmenter has normally just written the first keyframe of the segment to the analyzer. So if the analyzer tracks the last timecode scene, we should be able to use that as the segment's timecode. Right?

The behavior of timecode in the analyzer just seems a bit weird to me.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -176,6 +191,53 @@ impl Stream {
...
+                                            minutes: t.minutes.0,
+                                            seconds: t.seconds.0,
+                                            frames: t.n_frames.0,
+                                        };
+                                        if let Some(previous_timecode) = acc.iter().last() {
+                                            if t.full_timestamp_flag.0 == 0 {
+                                                if t.seconds_flag.0 == 0 {
+                                                    timecode.seconds = previous_timecode.seconds;
+                                                    timecode.minutes = previous_timecode.minutes;
+                                                    timecode.hours = previous_timecode.hours;
```
ccbrown, line 224: Can we find or make a video that exercises this code? My interpretation of the spec was that we should use the previous timecode's values, even if the previous timecode was for a previous access unit (i.e. not in the same SEI).

mpegts-segmenter/src/analyzer.rs
```diff
@@ -4,6 +4,14 @@ use std::error::Error;
 
 pub type Result<T> = std::result::Result<T, Box<dyn Error + Send + Sync>>;
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct StreamTimecode {
```
ccbrown, line 8: Could probably simplify to just `Timecode` especially since this will be used outside of the context of `Stream`.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -176,6 +195,57 @@ impl Stream {
                                 0 => 0.0,
                                 num_units_in_tick @ _ => (sps.vui_parameters.time_scale.0 as f64 / (2.0 * num_units_in_tick as f64) * 100.0).round() / 100.0,
                             };
+                            last_vui_parameters = Some(sps.vui_parameters);
+                        }
+                        h264::NAL_UNIT_TYPE_SUPPLEMENTAL_ENHANCEMENT_INFORMATION => {
+                            let bs = h264::Bitstream::new(nalu.iter().copied());
+                            let mut nalu = h264::NALUnit::decode(bs)?;
+
+                            if let Some(vui_params) = &last_vui_parameters {
```
ccbrown, line 204: As far as I can tell, `last_vui_parameters` is only `Some` when there was an SPS within the same access unit. Are SEI messages just ignored for non-keyframes?

mpegts-segmenter/src/analyzer.rs
```diff
@@ -241,6 +311,15 @@ impl Stream {
         Ok(())
     }
 
+    pub fn reset_timecode(&mut self) {
+        match self {
+            Self::AVCVideo { timecode_pending, .. } => {
+                *timecode_pending = true;
```
ccbrown, line 317: Could probably just set `timecode` to `None` and get rid of `timecode_pending`.

ccbrown (changes requested):
Okay, I still think this is doable without the resetting, but that's no big deal. This structure is just fine as long as we make sure it works properly.

Right now I don't believe it works properly for access units that don't also have SPSs (non-keyframes usually). I think everything else about it works fine, but as this is clearly tricky to get right, I'd really like to create some tests that 1. would confirm or disprove the above bug theory and 2. exercise the code for pic timing messages with full_timestamp_flag = 0.

mpegts-segmenter/src/analyzer.rs
dotcypress: But I need to store latest timecode in case current one has `full_timestamp_flag == 0`

mpegts-segmenter/src/analyzer.rs
dotcypress: Yes, this is a bug. 
Fixed

mpegts-segmenter/src/analyzer.rs
dotcypress: Agreed!

mpegts-segmenter/src/analyzer.rs
ccbrown: You're using `last_timecode` for that, which `timecode_pending` has no impact on.

dotcypress: Fixed!

ccbrown:
Yeah I think this looks good, but I do want to spend some time tomorrow looking at it more closely and trying to find some more test data.

ccbrown:
If you go here, a lot of the samples use full_timestamp_flag = 0:

https://www.mainconcept.com/us/getting-started/samples.html

You can use this one for example: http://downloads.mainconcept.com/MainConceptLogo_Blu-ray_AVC_1920x1080_LPCM.zip

Just use FFmpeg to convert / trim to a small .ts file we can commit.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -176,6 +193,56 @@ impl Stream {
...
+                                        h264::PicTiming::decode(&mut bs, &vui_params).unwrap().timecodes
+                                    })
+                                    .fold(acc, |mut acc, t| {
+                                        let mut timecode = Timecode {
+                                            hours: t.hours.0,
+                                            minutes: t.minutes.0,
+                                            seconds: t.seconds.0,
+                                            frames: t.n_frames.0,
+                                        };
+                                        if let Some(previous_timecode) = acc.iter().last() {
```
ccbrown, line 221: Isn't `acc` a `Vec`? Can't you just use `.last()` without iterating again?

```suggestion
                                        if let Some(previous_timecode) = acc.last() {
```

mpegts-segmenter/src/analyzer.rs
```diff
@@ -176,6 +193,56 @@ impl Stream {
...
+                                let mut rbsp = h264::Bitstream::new(&mut nalu.rbsp_byte);
+                                let sei = h264::SEI::decode(&mut rbsp)?;
+
+                                let acc: Vec<Timecode> = last_timecode.iter().cloned().collect();
+                                let timecodes = sei
+                                    .sei_message
+                                    .iter()
+                                    .flat_map(|sei_message| {
+                                        let mut bs = h264::Bitstream::new(sei_message.payload.iter().copied());
+                                        h264::PicTiming::decode(&mut bs, &vui_params).unwrap().timecodes
```
ccbrown, line 212: 🚨 🚓 🚨 `unwrap` isn't okay here.

ccbrown (changes requested):

mpegts-segmenter/src/analyzer.rs
```diff
@@ -176,6 +193,56 @@ impl Stream {
...
+                            if let Some(vui_params) = &last_vui_parameters {
+                                let mut rbsp = h264::Bitstream::new(&mut nalu.rbsp_byte);
+                                let sei = h264::SEI::decode(&mut rbsp)?;
+
+                                let acc: Vec<Timecode> = last_timecode.iter().cloned().collect();
+                                let timecodes = sei
+                                    .sei_message
+                                    .iter()
+                                    .flat_map(|sei_message| {
+                                        let mut bs = h264::Bitstream::new(sei_message.payload.iter().copied());
```
ccbrown, line 211: I think we can move out of sei_message here to simplify things:

```suggestion
                                    .sei_message
                                    .into_iter()
                                    .flat_map(|sei_message| {
                                        let mut bs = h264::Bitstream::new(sei_message.payload);
```

mpegts-segmenter/src/analyzer.rs
dotcypress: Oops!
Fixed

dotcypress:
> If you go here, a lot of the samples use full_timestamp_flag = 0:
> 
> https://www.mainconcept.com/us/getting-started/samples.html
> 
> You can use this one for example: http://downloads.mainconcept.com/MainConceptLogo_Blu-ray_AVC_1920x1080_LPCM.zip
> 
> Just use FFmpeg to convert / trim to a small .ts file we can commit.

New test added with video trimmed from 00:00:05 to 00:00:10, and resulting timecodes looks goods.

h264/src/sei.rs
```diff
@@ -90,7 +90,7 @@ impl PicTiming {
     pub fn decode<T: Iterator<Item = u8>>(bs: &mut Bitstream<T>, vui_params: &VUIParameters) -> io::Result<Self> {
         let mut ret = Self::default();
 
-        let hrd_params = vui_params.nal_hrd_parameters.as_ref().or(vui_params.vcl_hrd_parameters.as_ref());
+        let hrd_params = vui_params.nal_hrd_parameters.as_ref().or_else(||vui_params.vcl_hrd_parameters.as_ref());
```
ccbrown, line 93: What's this change for?

dotcypress: To disable Clippy warning

ccbrown (approved):

PR #32, opened 2020-11-09
Add Clippy support

dotcypress:
* Add FourCC constants
* Add Clippy to GitHub Actions and fix all major warnings


.github/workflows/rust.yml
```diff
@@ -24,14 +24,20 @@ jobs:
...
         override: true
     - name: Build
       uses: actions-rs/cargo@v1
       with:
         command: build
         args: --verbose
+    - name: Lint
+      uses: actions-rs/clippy@master
+      with:
+        args: --all-features --all-targets
```
ccbrown, line 37: Maybe we should add `-D warnings` so CI actually fails? This would require you to fix or `allow` the remaining warnings of course.

dotcypress: Done!

I have a bad news: Clippy Github action have a bug: workflow doesn't marked as failed when Clippy chgecks are failed.
https://github.com/actions-rs/clippy/issues/2

I think I can clone and fix this workflow later this week.

dotcypress: Let me fix all warnings first

ccbrown: Could we just invoke `cargo clippy` without using that action? I think we would loose the pretty annotations, but I think it's worth it to make CI do what it's supposed to do.

dotcypress: Yes, lets invoke just `cargo clippy` for now. 
I will send a new PR after github action bug fix.

ccbrown (approved):
Awesome! Thanks for cleaning all this up! Looks great!

PR #34, opened 2020-12-07
attempt to guess frame rates

ccbrown:
I think this is the most reasonable way to determine the frame rate of these 8K streams. This only impacts cases where the emitted frame rate would have previously been 0.0 due to missing parameter set timing info. For streams with precise timing, the guessed frame rate should be perfectly accurate. For streams where wall-clock time is used for presentation times, it's a bit of a gamble, but there's nothing more we can do.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.42.0
+        toolchain: 1.48.0
         components: clippy
         override: true
     - name: Build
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -18,12 +18,12 @@ jobs:
     - name: Set Up macOS
       if: startsWith(matrix.os, 'macos-')
       run: |
-        echo "::set-env name=LIBCLANG_PATH::/Library/Developer/CommandLineTools/usr/lib/libclang.dylib"
+        echo "LIBCLANG_PATH=/Library/Developer/CommandLineTools/usr/lib/libclang.dylib" >> $GITHUB_ENV
```
ccbrown, line 21: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

dotcypress (approved):
LGTM

PR #36, opened 2020-12-30
facilitate srt timestamping

ccbrown:
I want to avoid using PTSs for wallclock. It just makes playback too fragile.

Instead I want to use SRT timestamps, which in theory should be the same within a millisecond or two.

To facilitate that, this does two things:

1. Adds a `receive` method to `srt::AsyncStream` which does a receive and returns the source timestamp for the message.
2. Adds a `WriteContext` associated type to `SegmentStorage`. `Segmenter`'s `write` method now accepts an argument of this type, and this gets forwarded along to the segmenter when segments are ready for storage. This allows us to pass a timestamp through to storage.


sxing (approved):
Looks good, going to SRT_MSGCTRL.srctime instead of PTS. Will put discussion on what I need to do in Slack, but it looks pretty straightforward.

sxing (approved):

PR #37, opened 2021-01-10
support empty adaptation field

ccbrown:
Relevant spec:

Screen Shot 2021-01-10 at 4 37 07 PM

I didn't think zero length adaptation fields were valid but apparently they are. Some of our streams have these sometimes. Previously we would error out on them. Now we handle them correctly.


dskart:
not sure exactly what is happening here as I am not familiar enough with the nitty gritty of mpeg2, but don't see any rust related issues. I am approving so that you can deploy it

dskart (approved):

PR #40, opened 2021-04-02
add h265-tile-join program

ccbrown:
* Adds h265-tile-join, a demo program / library for joining untiled video into a single tiled video. See the conversation in Slack.
* Fixes `ceil_log2`, which I broke because `next_power_of_two` doesn't return the _next_ power of two for powers of two. Stupid name. :P Now it's fixed again and has more tests.
* Makes several things `Clone`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "h264",
     "h265",
     "h265-tile-mux",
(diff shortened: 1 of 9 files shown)


dskart (approved):
was gonna say you should make a test for `ceil_log2()` lol

PR #45, opened 2021-09-06
fix kCFStringEncodingUTF8 issue with newer bindgen

ccbrown:
Not sure how you guys have been working around this issue, but this should make it go away.

Newer bindgen versions were generating the type for this enum differently. This makes it always generate it the old way.


dotcypress (approved):
👍

PR #46, opened 2021-09-10
add insecure hmp client

ccbrown:
Allows connecting to insecure HMP servers. 😞

### DIFF
--- a/hmp/src/lib.rs
+++ b/hmp/src/lib.rs
@@ -24,6 +24,15 @@ impl Client {
         })
     }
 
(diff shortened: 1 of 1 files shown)


dotcypress (approved):
TLS is overrated 🤣

beiriannydd (approved):
Adds a "new_insecure" request method to create a client which ignores SSL Certificates.  There's danger in the method name.  I expect this to be called by the corresponding client

beiriannydd:
to be clear, this is still TLS encrypted, it just means that we aren't doing the usual validity checks to ensure the identity of the server we are connecting to.

PR #47, opened 2021-09-16
move adts parsing to mpeg4 crate

ccbrown:
Cuts ADTS parsing from the MPEGTS analyzer and pastes it into the MPEG4 crate for reusability.


dotcypress (approved):
👍

PR #52, opened 2021-11-08
update rust, add lint fixes

ccbrown:
This updates to Rust 1.56 and adds some lint fixes and exceptions with comments. I tried to add Trunk to CI, but ran into issues there, so this just resolves lint warnings.


dotcypress (approved):
🦀 🎉

PR #56, opened 2021-11-15
add packet encoding

ccbrown:
I think this is the last in my series of encoding PRs. This should cover everything I need to implement a muxer.


dotcypress (approved):
🦀

PR #59, opened 2022-02-11
Xilinx crate with decoder utalizes Xilinx API

epingel2:
### What it does 
Adds a module to av-rs containing the core utilities for Xilinx decoding. Within this module, this PR adds a decoder written in rust that leverages the Xilinx video-sdk APIs found [here](https://github.com/Xilinx/video-sdk) to take advantage of Xilinx's transcoding hardware.


xilinx/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+const XCLBIN_FILENAME: &str = "/opt/xilinx/xcdr/xclbins/transcode.xclbin";
+
+/// initalized the number of devices specified with the default xclbin_name
+///
+/// @decice_count: The number of devices available in the system.
+pub fn xlnx_init_all_devices(device_count: i32) -> Result<(), simple_error::SimpleError> {
+    let mut xclbin_params = Vec::new();
+
+    let fname = std::ffi::CString::new(XCLBIN_FILENAME).expect("CString::new failed");
+    let fname_ptr = fname.into_raw();
```
ccbrown, line 27: This looks like a memory leak:

> Consumes the CString and transfers ownership of the string to a C caller.
> 
> The pointer which this function returns must be returned to Rust and reconstituted using [CString::from_raw](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_raw) to be properly deallocated. Specifically, one should not use the standard C free() function to deallocate this string.
> 
> Failure to call [CString::from_raw](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_raw) will lead to a memory leak.

It should just be:

```suggestion
    let fname_ptr = fname.as_ptr();
```

xilinx/src/xlnx_dec_utils.rs
```diff
@@ -0,0 +1,239 @@
...
+    let mut buff_len = cu_pool_prop.cuListProp.cuProps[0].kernelName.len();
+    strcpy_to_arr_i8(&mut cu_pool_prop.cuListProp.cuProps[0].kernelName, buff_len, "decoder")?;
+    buff_len = cu_pool_prop.cuListProp.cuProps[0].kernelAlias.len();
+    strcpy_to_arr_i8(&mut cu_pool_prop.cuListProp.cuProps[0].kernelAlias, buff_len, "DECODER_MPSOC")?;
+    cu_pool_prop.cuListProp.cuProps[0].devExcl = false;
+    cu_pool_prop.cuListProp.cuProps[0].requestLoad = xrm_precision_1000000_bitmask(dec_load);
+
+    buff_len = cu_pool_prop.cuListProp.cuProps[0].kernelName.len();
+    strcpy_to_arr_i8(&mut cu_pool_prop.cuListProp.cuProps[1].kernelName, buff_len, "kernel_vcu_decoder")?;
+    buff_len = cu_pool_prop.cuListProp.cuProps[0].kernelAlias.len();
```
ccbrown, line 77: Copy/paste error?

```suggestion
    buff_len = cu_pool_prop.cuListProp.cuProps[1].kernelAlias.len();
```

xilinx/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    Ok(())
+}
+
+pub fn xrm_precision_1000000_bitmask(val: i32) -> i32 {
+    return val << 8;
+}
+
+// performs a strcpy from string literal to existing output array. ensures null termination.
+pub fn strcpy_to_arr_i8(buff: &mut [i8], buff_len: usize, in_str: &str) -> Result<(), simple_error::SimpleError> {
```
ccbrown, line 49: Is the `buff_len` argument necessary here? As far as I can tell, you could just use `buff.len()`.

xilinx/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    //check string length is smaller than buffer size. requires one byte for null termination
+    if in_str.len() > buff_len - 1 {
+        simple_error::bail!("Input str exceeds output buffer size")
+    }
+
+    for (i, c) in in_str.chars().enumerate() {
+        buff[i] = c as i8;
+    }
+    buff[in_str.len()] = '\0' as i8;
```
ccbrown, line 58: `as_bytes()` would also be good here:

```suggestion
    let src = in_str.as_bytes();
    buff[..src.len()].copy_from_slice(src);
    buff[src.len()] = 0;
```

xilinx/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        let result = 2 + 2;
+        assert_eq!(result, 4);
+    }
+}
```
ccbrown, line 69:
```suggestion
```

ccbrown:
Is there a good way to write some unit tests for this? Even if they require Xilinx hardware, it'd be good to be able to run tests manually.

epingel2:
> Is there a good way to write some unit tests for this? Even if they require Xilinx hardware, it'd be good to be able to run tests manually.

Yes I am working on that. When I get those in today i will change this from a pull request draft to being ready for review.

xilinx/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[dependencies]
+libloading = "0.7"
+simple-error = "0.2.1"
+
+[build-dependencies]
+# We're very permissive here with bindgen due to https://github.com/rust-lang/cargo/issues/5237
+bindgen = "0.*"
+
+[dev-dependencies]
+ffmpeg-sys = {version = "4.3.3", git = "https://github.com/meh/rust-ffmpeg-sys", default-features = false, features = ["avcodec", "avformat", "avfilter", "swscale"]}
```
ccbrown, line 17: It'd be nice to avoid the FFmpeg dependency here, even if it's just for development. I think we could do that and simplify the unit tests at the same time if we did something like this:

https://github.com/sportsball-ai/av-rs/blob/240a77b20b3e568aaf1083ab63c466661e5f8ec9/macos/video-toolbox/src/decompression_session.rs#L74

epingel2: The dependency on ffmpeg in testing have been removed and replaced with reading the test file directly.

ccbrown:
Since we can't easily test this in CI right now, we'll need to exclude this package from the tests. Does Cargo's `--exclude` flag make CI happy?

xilinx/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    Ok(())
+}
+
+pub fn xrm_precision_1000000_bitmask(val: i32) -> i32 {
+    val << 8
+}
+
+// performs a strcpy from string literal to existing output array. ensures null termination.
+pub fn strcpy_to_arr_i8(buf: &mut [i8], in_str: &str) -> Result<(), simple_error::SimpleError> {
```
ccbrown, line 46: This and several other things here probably shouldn't be public. `pub(crate)` maybe:

```suggestion
pub(crate) fn strcpy_to_arr_i8(buf: &mut [i8], in_str: &str) -> Result<(), simple_error::SimpleError> {
```

xilinx/src/xlnx_dec_utils.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    unsafe {
+        let ret = xrmExecPluginFunc(xrm_ctx, plugin_name, func_id, &mut plugin_param);
+        if ret != XRM_SUCCESS as i32 {
+            bail!("XRM decoder plugin failed to calculate decoder load. error: {}", ret);
+        }
+    }
+    // parse the load from the output buffer of plugin param.
+    let output_bytes = &plugin_param.output.map(|i| i as u8);
+    let dec_plugin_output = from_utf8(output_bytes).unwrap_or("-1").split(' ').collect::<Vec<&str>>()[0];
```
ccbrown, line 55: Maybe just use `next` instead of creating a `Vec`:

```suggestion
    let dec_plugin_output = from_utf8(output_bytes).unwrap_or("-1").split(' ').next().expect("split should emit at least one item");
```

xilinx/src/xlnx_dec_utils.rs
```diff
@@ -0,0 +1,228 @@
...
+        return Ok(());
+    }
+
+    let mut cu_pool_prop: xrmCuPoolProperty = Default::default();
+    xlnx_fill_dec_pool_props(&mut cu_pool_prop, xlnx_dec_ctx.dec_load)?;
+
+    unsafe {
+        let num_cu_pool = xrmCheckCuPoolAvailableNum(xlnx_dec_ctx.xrm_ctx, &mut cu_pool_prop);
+        if num_cu_pool <= 0 {
+            bail!("No decoder resources available for allocation")
```
ccbrown, line 96: Noticing inconsistent caps on these errors. Typically we use lowercase errors:

```suggestion
            bail!("no decoder resources available for allocation")
```

xilinx/src/xlnx_dec_utils.rs
```diff
@@ -0,0 +1,228 @@
...
+        xrmCuAllocFromDev(
+            xlnx_dec_ctx.xrm_ctx,
+            xlnx_dec_ctx.device_id,
+            &mut decode_cu_sw_prop,
+            &mut xlnx_dec_ctx.cu_list_res.cuResources[1],
+        )
+    };
+    if ret <= XRM_ERROR {
+        bail!("xrm failed to allocate decoder resources on device: {}", xlnx_dec_ctx.device_id);
+    }
```
ccbrown, line 144: It looks like this is repeated twice with a different `xlnx_dec_ctx.cu_list_res.cuResources` index. Could this just be a for loop?

xilinx/src/xlnx_decoder.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// @size: size of input data.
+    pub fn xlnx_dec_send_pkt(&mut self, buf: &mut [u8], pts: i32) -> Result<(), XlnxError> {
+        let mut data_used = 0;
+        let mut index = 0;
+        let mut ret;
+
+        let size = buf.len();
+        while index < size {
+            unsafe {
+                self.in_buf = xma_data_buffer_alloc(size as u64, false);
```
ccbrown, line 52: How does the previous buffer get released? Should we call `xma_data_buffer_free` first if `in_buf` is non-null?

xilinx/src/xlnx_dec_utils.rs
epingel2: They are very similar. however, one is using the `decode_cu_hw_prop` and the other uses the `decode_cu_sw_prop` so it cannot simply be done in a for-loop.

xilinx/src/xlnx_decoder.rs
epingel2: Thanks for pointing this out. The underlying data buffer gets deallocated by xvbm automatically when the frames get passed to the next plugin (scaling or encoding) however, using xma_data_buffer_alloc here was actually unnecessary since this is just an intermediate data structure used to send frames to the decider, it doesn't have to be allocated on the stack.

xilinx/src/xlnx_decoder.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        let dec_session = xlnx_create_dec_session(xma_dec_props, xlnx_dec_ctx)?;
+
+        let mut frame_props: XmaFrameProperties = Default::default();
+        let ret = unsafe { xma_dec_session_get_properties(dec_session, &mut frame_props) };
+        if ret != XMA_SUCCESS as i32 {
+            bail!("unable to get frame properties from decoder session")
+        }
+
+        let in_buf: XmaDataBuffer = Default::default();//unsafe { xma_data_buffer_alloc(0, true) };
```
ccbrown, line 27: Is this commented out code needed?

```suggestion
        let in_buf: XmaDataBuffer = Default::default();
```

ccbrown (commented):
It's really hard to follow the ownership of the C resources here. How confident are you that there are no leaked resources? Is there any way to increase that confidence with more RAII?

ccbrown (approved):

PR #66, opened 2022-03-16
Xilinx scaler

epingel2:
This implements the Xilinx scalar API. This pull request builds off of work done in the Xilinx-decoder branch.
Notable changes include moving all of the unit tests to the lib.rs file. This was done because initialization of the Xilinx needs to occur once and only once for a given application, and this allows the unit tests to share the resources without causing problems for each other.

Steps to test:
Begin with a server or machine with the xilinx u30 acceleration cards.
- install xilinx video sdk.
- run xilinx setup script : `source /opt/xilinx/xcdr/setup.sh`
- set env `LD_LIBRARY_PATH=/opt/xilinx/xrm/lib:/opt/xilinx/xrt/lib`
- `cargo test`


xilinx/src/xlnx_scaler.rs
```diff
@@ -0,0 +1,112 @@
+use crate::{sys::*, xlnx_error::*, xlnx_scal_utils::*};
+use alloc::vec::Vec;
```
ccbrown, line 2: Is this necessary? This isn't a no_std crate, so Vec is automatically included from the std prelude right?

xilinx/src/lib.rs
```diff
@@ -55,3 +68,316 @@ pub(crate) fn strcpy_to_arr_i8(buf: &mut [i8], in_str: &str) -> Result<(), simpl
...
+
+    static INIT: Once = Once::new();
+
+    pub fn initialize() {
+        INIT.call_once(|| {
+            xlnx_init_all_devices(2).unwrap();
+        });
+    }
+
+    fn decode_file(file_path: &str, codec_type: u32, profile: u32, level: u32) -> (i32, i32) {
```
ccbrown, line 93: Files in a crate can use functions from other files, even if they're in the tests module. You just need to add `pub` to `mod tests {`. So it shouldn't be a problem to keep this with the decoder implementation.

ccbrown (commented):
I think this looks good. Two minor comments though.

xilinx/src/xlnx_scaler.rs
epingel2: Yes, I will remove this.

xilinx/src/lib.rs
epingel2: Thanks for the suggestion. I have figured out how to split out the test cases for the scalar and the decoder into their respective files, and use the lib test module to hold the initialization code state for Xilinx.

xilinx/src/xlnx_scaler.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn drop(&mut self) {
+        unsafe {
+            if !self.scal_session.is_null() {
+                xma_scaler_session_destroy(self.scal_session);
+            }
+            for i in 0..SCAL_MAX_ABR_CHANNELS {
+                if !self.out_frame_list[i].is_null() {
+                    xma_frame_free(self.out_frame_list[i]);
+                }
+            }
```
ccbrown, line 108: Could you simplify to this instead?

```suggestion
            for f in &self.out_frame_list {
                if !f.is_null() {
                    xma_frame_free(f);
                }
            }
```

ccbrown (approved):

PR #67, opened 2022-03-29
fix x264 encoding in release mode

ccbrown:
The x264 unit test was passing in debug mode, but failing in release mode. This fixes it.

### DIFF
--- a/x264/src/encoder.rs
+++ b/x264/src/encoder.rs
@@ -88,7 +88,7 @@ impl<F> X264Encoder<F> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


x264/src/encoder.rs
```diff
@@ -101,8 +101,8 @@ impl<F> X264Encoder<F> {
                 self.encoder,
                 &mut nals as _,
                 &mut nal_count as _,
-                match pic {
-                    Some(mut pic) => &mut pic as _,
```
ccbrown, line 105: `pic` was being dropped at the end of this statement, before x264 had a chance to read it.

epingel2 (approved):

PR #68, opened 2022-03-30
Adds encoder to xilinx crate

epingel2:
This implements the Xilinx encoder API.

Steps to test:
Begin with a server or machine with the xilinx u30 acceleration cards.

- install xilinx video sdk.
- run xilinx setup script : source /opt/xilinx/xcdr/setup.sh
- set env LD_LIBRARY_PATH=/opt/xilinx/xrm/lib:/opt/xilinx/xrt/lib
- cargo test

### DIFF
--- a/xilinx/src/xlnx_enc_props.rs
+++ b/xilinx/src/xlnx_enc_props.rs
@@ -1,0 +1,372 @@
+use simple_error::{bail, SimpleError};
+
+use crate::{strcpy_to_arr_i8, sys::*};
+
+pub const MAX_ENC_PARAMS: usize = 3;
+
+const ENC_OPTIONS_PARAM_NAME: &[u8] = b"enc_options\0";
+const LATENCY_LOGGING_PARAM_NAME: &[u8] = b"latency_logging\0";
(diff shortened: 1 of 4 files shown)


xilinx/src/xlnx_enc_props.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    if enc_props.tune_metrics {
+        scaling_list = "FLAT";
+        qp_ctrl_mode = "UNIFORM_QP";
+    }
+
+    let enc_options;
+    // hevc
+    if enc_props.codec_id == CODEC_ID_HEVC {
+        enc_options = format!("[INPUT]\n\
```
ccbrown, line 224: I think you could make these huge formatted strings much better with named parameters, introduced in Rust 1.58:

https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings

epingel2: That's a great idea. I didn't know about this before. Looks a lot cleaner.

ccbrown (approved):
👍

PR #70, opened 2022-04-06
add video toolbox encoder

ccbrown:
Adds two new types to the video toolbox crate:

* An unopinionated abstraction over [VTCompressionSession](https://developer.apple.com/documentation/videotoolbox/vtcompressionsession?language=objc).
* An implementation of `av_traits::VideoEncoder`.

This enables encoding video as H264 or H265 using hardware acceleration on macOS.

### DIFF
--- a/macos/core-foundation/src/dictionary.rs
+++ b/macos/core-foundation/src/dictionary.rs
@@ -1,5 +1,4 @@
 use super::*;
-use crate as core_foundation;
 use std::os::raw::c_void;
 
 pub struct Dictionary(sys::CFDictionaryRef);
--- a/macos/core-foundation/src/dictionary.rs
+++ b/macos/core-foundation/src/dictionary.rs
@@ -1,0 +1,49 @@
+use super::*;
+use crate as core_foundation;
+use std::os::raw::c_void;
+
+pub struct Dictionary(sys::CFDictionaryRef);
+crate::trait_impls!(Dictionary);
+
+impl Dictionary {
+    /// # Safety
+    /// Behavior is undefined if the value is not of type `T`.
+    pub unsafe fn cf_type_value<T: CFType>(&self, key: *const c_void) -> Option<T> {
+        let mut v = std::ptr::null();
+        if sys::CFDictionaryGetValueIfPresent(self.0, key, &mut v as *mut _ as _) != 0 {
+            Some(T::with_cf_type_ref(v as _))
+        } else {
+            None
+        }
+    }
+}
+
+pub struct MutableDictionary(sys::CFMutableDictionaryRef);
+crate::trait_impls!(MutableDictionary);
+
+impl Default for MutableDictionary {
+    fn default() -> Self {
+        unsafe {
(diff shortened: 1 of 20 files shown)


macos/core-foundation/src/dictionary.rs
```diff
@@ -0,0 +1,48 @@
...
+
+pub struct MutableDictionary(sys::CFMutableDictionaryRef);
+crate::trait_impls!(MutableDictionary);
+
+impl Default for MutableDictionary {
+    fn default() -> Self {
+        unsafe {
+            Self(sys::CFDictionaryCreateMutable(
+                std::ptr::null_mut(),
+                0,
```
dskart, line 28: where does this 0 come from

macos/core-foundation/src/lib.rs
```diff
@@ -15,31 +15,40 @@ pub mod sys {
...
+pub use array::*;
+
+pub mod boolean;
+pub use boolean::*;
+
+pub mod dictionary;
+pub use dictionary::*;
+
 #[macro_export]
 macro_rules! trait_impls {
```
dskart, line 28: wouldn't using a procedural "derive" macro be better for this? so you can do `#[derive(MySpecialTraits)]`

dskart: [https://doc.rust-lang.org/book/ch19-06-macros.html#how-to-write-a-custom-derive-macro](https://doc.rust-lang.org/book/ch19-06-macros.html#how-to-write-a-custom-derive-macro)

dskart (changes requested):
I think this looks fine, I have a few questions in comments.

I will say the core-* folders are missing tests and it makes it hard as an external reader to really validate if they are going to work or not.

Especially all the type definitions and unsafe blocks I kinda have to take your "word" for it.

Maybe this is too much work and not possible with the nature of those types but it would be good to have simple tests for all of those.

macos/core-foundation/src/lib.rs
ccbrown: Probably. I think that's a task for another PR though.

ccbrown:
> I think this looks fine, I have a few questions in comments.
> 
> I will say the core-* folders are missing tests and it makes it hard as an external reader to really validate if they are going to work or not.
> 
> Especially all the type definitions and unsafe blocks I kinda have to take your "word" for it.
> 
> Maybe this is too much work and not possible with the nature of those types but it would be good to have simple tests for all of those.

It's definitely possible, and I'm adding a few now, but the real challenge is that the types are incomplete. I'm implementing functionality as-needed, and don't even have functionality for creating and populating an array for example. So adding good tests would require that first.

On the upside, all of these are really light wrappers around C functions, so the possibility for error is low.

macos/core-foundation/src/dictionary.rs
ccbrown: The "capacity" parameter. I just added a `with_capacity` function similar to `Vec::with_capacity` to make this clearer and more useful.

dskart (approved):

PR #75, opened 2022-04-22
Init now query devices. Made some fields pubic

epingel2:
What it does: the initialize function now queries the number of devices instead of taking a number as a parameter. This allows us to attempt to initialize and query if there are devices on the system or not.

There were also some fields that needed to be changed (made public and changing types)


xilinx/src/xlnx_encoder.rs
```diff
@@ -236,7 +236,7 @@ mod encoder_tests {
 
         processed_frame_count
     }
-
+
```
ccbrown, line 239:
```suggestion
```

ccbrown (approved):

lzuosym (approved):

PR #82, opened 2022-04-28
Fix encoding for timestamps PTS and DTS in PES PacketHeader

lzuosym:


ccbrown (approved):
Wow, really good catch! I guess there's a good chance that this has never impacted us since it requires the timestamps to be really high, but this is a super important fix.

lzuosym:
Yep!

PR #83, opened 2022-05-03
Create a new muxer to buffer the packets internally as needed

lzuosym:
Ensure that the packets written are properly interleaved in the order of increasing timestamp.


lzuosym:
@ccbrown I tried to re-use and wrap the existing `Muxer` in the new `InterleavingMuxer`, but I failed to figured out a way to do so, since rust compiler won't allow me to reference and modify the content of a `Stream` from a different `Stream`. 
```
impl<'a, W: Write> Stream<'a, W> {
    pub fn write(&mut self, p: Packet) -> Result<(), EncodeError> {
       //  Cannot emit buffered Packets in another Stream in here.
    }
}
```

Also not sure how to fully test this new Muxer.

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,354 @@
...
+                }
+            }
+        }
+        for (packet_id, packet) in packets {
+            self.write_packet(packet_id as usize, packet)?;
+        }
+        Ok(())
+    }
+
+    fn write_ts_packet(&mut self, p: ts::Packet) -> Result<(), EncodeError> {
```
ccbrown, line 185: Is there any difference between this function and `muxer::Muxer::write`?

Could we avoid duplicating this code by using `muxer::Muxer` without `muxer::Stream` and making `muxer::Muxer::write` `pub(crate)` so we can call it from this file?

lzuosym: We can reuse `muxer::Muxer::write` if we wrap all the variables in `InterleavingMuxer` to a `RefCell`, but there is no need for this extra layer of wrapper. If you compare the code, the above function is actually replaced all `state` with `self`, and function signature is changed from `&self` to `&mut self`.

ccbrown: It seems like there are a few ways you could make this work without duplicating large amounts of code. I don't think `RefCell` is heavy enough that the extra layer should be a concern, but you can easily avoid it by doing something like this:

```diff
diff --git a/mpeg2/src/muxer.rs b/mpeg2/src/muxer.rs
index e535242..dcde1f5 100644
--- a/mpeg2/src/muxer.rs
+++ b/mpeg2/src/muxer.rs
@@ -9,7 +9,7 @@ struct StreamState {
     data: Vec<u8>,
 }
 
-struct State<W> {
+pub(crate) struct State<W> {
     w: W,
     next_packet_id: u16,
     pcr_pid: u16,
@@ -70,6 +70,10 @@ impl<W: Write> Muxer<W> {
 
     fn write(&self, p: ts::Packet) -> Result<(), EncodeError> {
         let mut state = self.state.borrow_mut();
+        Self::do_write(&mut state, p)
+    }
+
+    pub(crate) fn do_write(state: &mut State<W>, p: ts::Packet) -> Result<(), EncodeError> {
         let pcr = p.adaptation_field.as_ref().and_then(|af| af.program_clock_reference_27mhz);
 
         let mut should_write_headers = !state.did_write_headers;

```

Then you could re-use `State` and `do_write` from here.

I think I'd really prefer to just refactor `Muxer` a little bit to support what you need though. We don't have to worry about backwards compatibility here. We could get rid of the stream lifetime and the interior mutability:

```diff
diff --git a/mpeg2/src/muxer.rs b/mpeg2/src/muxer.rs
index e535242..0b6e591 100644
--- a/mpeg2/src/muxer.rs
+++ b/mpeg2/src/muxer.rs
@@ -1,6 +1,5 @@
 use super::{pes, ts, EncodeError};
 use alloc::vec::Vec;
-use core::cell::RefCell;
 use core2::io::Write;
 
 struct StreamState {
@@ -9,7 +8,7 @@ struct StreamState {
     data: Vec<u8>,
 }
 
-struct State<W> {
+pub struct Muxer<W> {
     w: W,
     next_packet_id: u16,
     pcr_pid: u16,
@@ -19,10 +18,6 @@ struct State<W> {
     streams: Vec<StreamState>,
 }
 
-pub struct Muxer<W> {
-    state: RefCell<State<W>>,
-}
-
 #[derive(Clone, Debug)]
 pub struct StreamConfig {
     pub stream_id: u8,
@@ -37,30 +32,26 @@ pub struct StreamConfig {
 impl<W: Write> Muxer<W> {
     pub fn new(w: W) -> Self {
         Self {
-            state: RefCell::new(State {
-                w,
-                next_packet_id: 0x100,
-                pcr_pid: 0,
-                last_header_pcr: None,
-                did_write_headers: false,
-                header_continuity_counter: 0,
-                streams: vec![],
-            }),
+            w,
+            next_packet_id: 0x100,
+            pcr_pid: 0,
+            last_header_pcr: None,
+            did_write_headers: false,
+            header_continuity_counter: 0,
+            streams: vec![],
         }
     }
 
-    pub fn new_stream(&self, config: StreamConfig) -> Stream<'_, W> {
-        let mut state = self.state.borrow_mut();
-        let packet_id = state.next_packet_id;
-        state.next_packet_id += 1;
-        state.streams.push(StreamState {
+    pub fn new_stream(&mut self, config: StreamConfig) -> Stream {
+        let packet_id = self.next_packet_id;
+        self.next_packet_id += 1;
+        self.streams.push(StreamState {
             stream_type: config.stream_type,
             pid: packet_id,
             data: config.data,
         });
-        state.pcr_pid = packet_id;
+        self.pcr_pid = packet_id;
         Stream {
-            muxer: self,
             packet_id,
             continuity_counter: 0,
             stream_id: config.stream_id,
@@ -68,18 +59,43 @@ impl<W: Write> Muxer<W> {
         }
     }
 
-    fn write(&self, p: ts::Packet) -> Result<(), EncodeError> {
-        let mut state = self.state.borrow_mut();
+    pub fn write(&mut self, stream: &mut Stream, p: Packet) -> Result<(), EncodeError> {
+        let pes_packet = pes::Packet {
+            header: pes::PacketHeader {
+                stream_id: stream.stream_id,
+                optional_header: Some(pes::OptionalHeader {
+                    data_alignment_indicator: true,
+                    pts: p.pts_90khz,
+                    dts: p.dts_90khz,
+                }),
+                data_length: if stream.unbounded_data_length { 0 } else { p.data.len() },
+            },
+            data: p.data.into(),
+        };
+        for ts_packet in pes_packet.packetize(pes::PacketizationConfig {
+            packet_id: stream.packet_id,
+            continuity_counter: stream.continuity_counter,
+            random_access_indicator: p.random_access_indicator,
+        }) {
+            if ts_packet.payload.is_some() {
+                stream.continuity_counter = (stream.continuity_counter + 1) % 16;
+            }
+            self.write_ts_packet(ts_packet)?;
+        }
+        Ok(())
+    }
+
+    fn write_ts_packet(&mut self, p: ts::Packet) -> Result<(), EncodeError> {
         let pcr = p.adaptation_field.as_ref().and_then(|af| af.program_clock_reference_27mhz);
 
-        let mut should_write_headers = !state.did_write_headers;
-        if p.packet_id == state.pcr_pid && state.did_write_headers {
+        let mut should_write_headers = !self.did_write_headers;
+        if p.packet_id == self.pcr_pid && self.did_write_headers {
             // figure out if we need to repeat the headers
             if let Some(pcr) = pcr {
-                if state.last_header_pcr.is_none() {
-                    state.last_header_pcr = Some(pcr);
+                if self.last_header_pcr.is_none() {
+                    self.last_header_pcr = Some(pcr);
                 }
-                if let Some(last_pcr) = state.last_header_pcr {
+                if let Some(last_pcr) = self.last_header_pcr {
                     // repeat the headers if we've wrapped around or if 90ms has passed
                     if pcr + 27_000_000 < last_pcr || pcr > last_pcr + 2_430_000 {
                         should_write_headers = true;
@@ -117,18 +133,18 @@ impl<W: Write> Muxer<W> {
                 let p = ts::Packet {
                     packet_id: ts::PID_PAT,
                     payload_unit_start_indicator: true,
-                    continuity_counter: state.header_continuity_counter,
+                    continuity_counter: self.header_continuity_counter,
                     adaptation_field: None,
                     payload: Some(encoded.into()),
                 };
-                p.encode(&mut state.w)?;
+                p.encode(&mut self.w)?;
             }
 
             // write the PMT
             {
                 let pmt = ts::PMTData {
-                    pcr_pid: state.pcr_pid,
-                    elementary_stream_info: state
+                    pcr_pid: self.pcr_pid,
+                    elementary_stream_info: self
                         .streams
                         .iter()
                         .map(|s| ts::PMTElementaryStreamInfo {
@@ -156,19 +172,19 @@ impl<W: Write> Muxer<W> {
                 let p = ts::Packet {
                     packet_id: PMT_PID,
                     payload_unit_start_indicator: true,
-                    continuity_counter: state.header_continuity_counter,
+                    continuity_counter: self.header_continuity_counter,
                     adaptation_field: None,
                     payload: Some(encoded.into()),
                 };
-                p.encode(&mut state.w)?;
+                p.encode(&mut self.w)?;
             }
 
-            state.did_write_headers = true;
-            state.last_header_pcr = pcr;
-            state.header_continuity_counter = (state.header_continuity_counter + 1) % 16;
+            self.did_write_headers = true;
+            self.last_header_pcr = pcr;
+            self.header_continuity_counter = (self.header_continuity_counter + 1) % 16;
         }
 
-        p.encode(&mut state.w)?;
+        p.encode(&mut self.w)?;
         Ok(())
     }
 }
@@ -181,42 +197,13 @@ pub struct Packet<'a> {
     pub dts_90khz: Option<u64>,
 }
 
-pub struct Stream<'a, W> {
-    muxer: &'a Muxer<W>,
+pub struct Stream {
     packet_id: u16,
     continuity_counter: u8,
     stream_id: u8,
     unbounded_data_length: bool,
 }
 
-impl<'a, W: Write> Stream<'a, W> {
-    pub fn write(&mut self, p: Packet) -> Result<(), EncodeError> {
-        let pes_packet = pes::Packet {
-            header: pes::PacketHeader {
-                stream_id: self.stream_id,
-                optional_header: Some(pes::OptionalHeader {
-                    data_alignment_indicator: true,
-                    pts: p.pts_90khz,
-                    dts: p.dts_90khz,
-                }),
-                data_length: if self.unbounded_data_length { 0 } else { p.data.len() },
-            },
-            data: p.data.into(),
-        };
-        for ts_packet in pes_packet.packetize(pes::PacketizationConfig {
-            packet_id: self.packet_id,
-            continuity_counter: self.continuity_counter,
-            random_access_indicator: p.random_access_indicator,
-        }) {
-            if ts_packet.payload.is_some() {
-                self.continuity_counter = (self.continuity_counter + 1) % 16;
-            }
-            self.muxer.write(ts_packet)?;
-        }
-        Ok(())
-    }
-}
-
 #[cfg(test)]
 mod test {
     use super::*;
@@ -235,7 +222,7 @@ mod test {
 
         let mut data_out = Vec::new();
         {
-            let muxer = Muxer::new(&mut data_out);
+            let mut muxer = Muxer::new(&mut data_out);
             let mut video_out = muxer.new_stream(StreamConfig {
                 stream_id: 0xe0,
                 stream_type: 0x1b,
@@ -250,13 +237,16 @@ mod test {
                 }
 
                 for frame in video_in.write(&p).unwrap() {
-                    video_out
-                        .write(Packet {
-                            data: &frame.data,
-                            random_access_indicator,
-                            pts_90khz: frame.header.optional_header.as_ref().and_then(|h| h.pts),
-                            dts_90khz: frame.header.optional_header.as_ref().and_then(|h| h.dts),
-                        })
+                    muxer
+                        .write(
+                            &mut video_out,
+                            Packet {
+                                data: &frame.data,
+                                random_access_indicator,
+                                pts_90khz: frame.header.optional_header.as_ref().and_then(|h| h.pts),
+                                dts_90khz: frame.header.optional_header.as_ref().and_then(|h| h.dts),
+                            },
+                        )
                         .unwrap();
                     random_access_indicator = false;
                 }
```

lzuosym: 👍 it helps a lot to simplify the code. A few more changes in `muxer`, added `pub(crate)` to all the variables in `pub struct Muxer<W>` and `pub(crate) struct StreamState`. And removed the lifetime in `Packet`
```
pub struct Packet {
    pub data: Vec<u8>,
```

ccbrown: Looks like you could also get rid of all the `pub(crate)`s while eliminating more duplicate code:

```
diff --git a/mpeg2/src/interleaving_muxer.rs b/mpeg2/src/interleaving_muxer.rs
index 8c08f89..efd13ad 100644
--- a/mpeg2/src/interleaving_muxer.rs
+++ b/mpeg2/src/interleaving_muxer.rs
@@ -1,5 +1,5 @@
 use super::EncodeError;
-use crate::muxer::{Muxer, Packet, Stream, StreamConfig, StreamState};
+use crate::muxer::{Muxer, Packet, Stream, StreamConfig};
 use alloc::vec::Vec;
 use core2::io::Write;
 use std::collections::VecDeque;
@@ -20,15 +20,7 @@ pub struct InterleavingStream {
 
 impl<W: Write> InterleavingMuxer<W> {
     pub fn new(w: W, max_buffer_duration: Duration) -> Self {
-        let muxer = Muxer {
-            w,
-            next_packet_id: 0x100,
-            pcr_pid: 0,
-            last_header_pcr: None,
-            did_write_headers: false,
-            header_continuity_counter: 0,
-            streams: vec![],
-        };
+        let muxer = Muxer::new(w);
         Self {
             muxer,
             max_buffer_duration: (max_buffer_duration.as_millis() * 90) as u64,
@@ -38,23 +30,9 @@ impl<W: Write> InterleavingMuxer<W> {
     }
 
     pub fn add_stream(&mut self, config: StreamConfig) {
-        let packet_id = self.muxer.next_packet_id;
-        let stream_state = StreamState {
-            stream_type: config.stream_type,
-            pid: packet_id,
-            data: config.data,
-        };
-        self.muxer.next_packet_id += 1;
-        self.muxer.streams.push(stream_state);
-        self.muxer.pcr_pid = packet_id;
-        let internal_stream = Stream {
-            packet_id,
-            continuity_counter: 0,
-            stream_id: config.stream_id,
-            unbounded_data_length: config.unbounded_data_length,
-        };
+        let wrapper = self.muxer.new_stream(config);
         let interleaving_stream = InterleavingStream {
-            wrapper: internal_stream,
+            wrapper,
             buffered_packets: VecDeque::with_capacity(6),
             last_written_ts: 0,
         };
@@ -97,7 +75,7 @@ impl<W: Write> InterleavingMuxer<W> {
                     }
                 })
             {
-                let min_ts = self.min_ts_excluding(self.streams[index].wrapper.packet_id as usize);
+                let min_ts = self.min_ts_excluding(self.streams[index].wrapper.packet_id() as usize);
                 let stream = &mut self.streams[index];
                 min_ts_stream_index = index;
                 while let Some(p) = stream.buffered_packets.front() {
@@ -132,7 +110,7 @@ impl<W: Write> InterleavingMuxer<W> {
             while let Some(p) = stream.buffered_packets.front() {
                 let ts = p.dts_90khz.or(p.pts_90khz).unwrap_or(stream.last_written_ts);
                 if self.largest_ts_in_buffer - ts > self.max_buffer_duration {
-                    packets.push((stream.wrapper.packet_id, stream.buffered_packets.pop_front().unwrap()));
+                    packets.push((stream.wrapper.packet_id(), stream.buffered_packets.pop_front().unwrap()));
                 }
             }
         }
@@ -144,7 +122,7 @@ impl<W: Write> InterleavingMuxer<W> {
 
     fn min_ts_excluding(&self, packet_id: usize) -> u64 {
         let mut min_ts = u64::MAX;
-        for s in self.streams.iter().filter(|s| s.wrapper.packet_id as usize != packet_id) {
+        for s in self.streams.iter().filter(|s| s.wrapper.packet_id() as usize != packet_id) {
             if let Some(p) = s.buffered_packets.front() {
                 min_ts = min_ts.min(p.dts_90khz.or(p.pts_90khz).unwrap_or(0));
             }
diff --git a/mpeg2/src/muxer.rs b/mpeg2/src/muxer.rs
index 9cf0652..3817835 100644
--- a/mpeg2/src/muxer.rs
+++ b/mpeg2/src/muxer.rs
@@ -9,13 +9,13 @@ pub(crate) struct StreamState {
 }
 
 pub struct Muxer<W> {
-    pub(crate) w: W,
-    pub(crate) next_packet_id: u16,
-    pub(crate) pcr_pid: u16,
-    pub(crate) did_write_headers: bool,
-    pub(crate) last_header_pcr: Option<u64>,
-    pub(crate) header_continuity_counter: u8,
-    pub(crate) streams: Vec<StreamState>,
+    w: W,
+    next_packet_id: u16,
+    pcr_pid: u16,
+    did_write_headers: bool,
+    last_header_pcr: Option<u64>,
+    header_continuity_counter: u8,
+    streams: Vec<StreamState>,
 }
 
 #[derive(Clone, Debug)]
@@ -198,10 +198,16 @@ pub struct Packet {
 }
 
 pub struct Stream {
-    pub(crate) packet_id: u16,
-    pub(crate) continuity_counter: u8,
-    pub(crate) stream_id: u8,
-    pub(crate) unbounded_data_length: bool,
+    packet_id: u16,
+    continuity_counter: u8,
+    stream_id: u8,
+    unbounded_data_length: bool,
+}
+
+impl Stream {
+    pub fn packet_id(&self) -> u16 {
+        self.packet_id
+    }
 }
 
 #[cfg(test)]
```

Simplifying this and boiling the logic in this new file down to only the interleaving functionality will make it much easier for me to review this.

lzuosym: forgot to call the constructors. :disappointed:

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,189 @@
+use super::EncodeError;
+use crate::muxer::{Muxer, Packet, Stream, StreamConfig};
+use alloc::vec::Vec;
+use core2::io::Write;
+use std::collections::VecDeque;
+use std::time::Duration;
+
+pub struct InterleavingMuxer<W> {
+    muxer: Muxer<W>,
+    max_buffer_duration: u64,
```
ccbrown, line 10: Can we document the units here somehow? A comment would be alright, but I would probably just put it in the field name:

```suggestion
    max_buffer_duration_90khz: u64,
```

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,189 @@
...
+        Ok(())
+    }
+
+    fn write_packets_outside_buffer_duration(&mut self) -> Result<(), EncodeError> {
+        let mut packets = vec![];
+        for (stream_index, stream) in self.streams.iter_mut().enumerate() {
+            while let Some(p) = stream.buffered_packets.front() {
+                let ts = p.dts_90khz.or(p.pts_90khz).unwrap_or(stream.last_written_ts);
+                if self.largest_ts_in_buffer - ts > self.max_buffer_duration {
+                    packets.push((stream_index, stream.buffered_packets.pop_front().unwrap()));
```
ccbrown, line 113: Why not just call `self.write_packet` here?

Also, prefer `expect` over `unwrap`:

```suggestion
                    packets.push((stream_index, stream.buffered_packets.pop_front().expect("there must be at least one packet to pop")));
```

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,189 @@
...
+                    random_access_indicator = false;
+                }
+            }
+        }
+
+        assert!(data_out.len() >= 8000000);
+
+        // uncomment this to write to a file for testing with other programs
+        // File::create("tmp.ts").unwrap().write_all(&data_out).unwrap();
+    }
```
ccbrown, line 188: This is a fine baseline test, but it doesn't actually test the interleaving functionality. Before approving, I need to be convinced by the unit tests that interleaving works as intended.

Can we add some unit tests to cover cases like:

* With a single stream, the packets should be passed through with zero buffering (and zero copying of the data).
* With multiple out-of-sync streams, packets should be interleaved like we discussed.
* If stream A stops and stream B continues, all packets from stream A should eventually be written, and all packets from stream B should be written after the max buffer duration.
* If the timestamps roll over, interleaving should work without any disruption.
* No data should be lost at the end of the streams, when the muxer is dropped.

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn min_ts_excluding(&self, packet_id: usize) -> u64 {
+        let mut min_ts = u64::MAX;
+        for s in self.streams.iter().filter(|s| s.wrapper.packet_id() as usize != packet_id) {
+            if let Some(p) = s.buffered_packets.front() {
+                min_ts = min_ts.min(p.dts_90khz.or(p.pts_90khz).unwrap_or(0));
+            }
+        }
+        min_ts
+    }
+}
```
ccbrown, line 132: Can we add a `flush()` function to signal to the muxer that there won't be any new data, and any buffered data should be written out, in order? This should be automatically called as part of a `Drop` implementation. We can't have data get lost at the end of the streams.

ccbrown (changes requested):
Alright, I like the way this looks now. I have a few minor comments here and one big one on unit tests.

mpeg2/src/muxer.rs
```diff
@@ -156,64 +172,41 @@ impl<W: Write> Muxer<W> {
...
+        p.encode(&mut self.w)?;
         Ok(())
     }
 }
 
 #[derive(Default)]
-pub struct Packet<'a> {
-    pub data: &'a [u8],
+pub struct Packet {
+    pub data: Vec<u8>,
```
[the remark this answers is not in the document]
lzuosym: If adding a lifetime to Packet, we seems to be not able to buffer this Packet in `InterleavingStream`. Data are moved into this Packet without lifetime, no copies. But it depends on how clients call it.


```
error[E0597]: `frame.data` does not live long enough
   --> mpeg2/src/interleaving_muxer.rs:172:39
    |
168 | /                     interleaving_muxer
169 | |                         .write(
170 | |                             0,
171 | |                             Packet {
172 | |                                 data: &frame.data,
    | |                                       ^^^^^^^^^^^ borrowed value does not live long enough
...   |
176 | |                             },
177 | |                         )
    | |_________________________- borrow later used here
...
180 |                   }
    |                   - `frame.data` dropped here while still borrowed
```


```diff
diff --git a/mpeg2/src/interleaving_muxer.rs b/mpeg2/src/interleaving_muxer.rs
index 10afa3f..28464bf 100644
--- a/mpeg2/src/interleaving_muxer.rs
+++ b/mpeg2/src/interleaving_muxer.rs
@@ -5,20 +5,20 @@ use core2::io::Write;
 use std::collections::VecDeque;
 use std::time::Duration;

-pub struct InterleavingMuxer<W> {
+pub struct InterleavingMuxer<'a, W> {
     muxer: Muxer<W>,
     max_buffer_duration: u64,
     largest_ts_in_buffer: u64,
-    streams: Vec<InterleavingStream>,
+    streams: Vec<InterleavingStream<'a>>,
 }

-pub struct InterleavingStream {
+pub struct InterleavingStream<'a> {
     wrapper: Stream,
-    buffered_packets: VecDeque<Packet>,
+    buffered_packets: VecDeque<Packet<'a>>,
     last_written_ts: u64,
 }

-impl<W: Write> InterleavingMuxer<W> {
+impl<'a, W: Write> InterleavingMuxer<'a, W> {
     pub fn new(w: W, max_buffer_duration: Duration) -> Self {
         let muxer = Muxer::new(w);
         Self {
@@ -39,7 +39,7 @@ impl<W: Write> InterleavingMuxer<W> {
         self.streams.push(interleaving_stream);
     }

-    pub fn write(&mut self, stream_index: usize, p: Packet) -> Result<(), EncodeError> {
+    pub fn write(&mut self, stream_index: usize, p: Packet<'a>) -> Result<(), EncodeError> {
         let last_written_ts = self.streams[stream_index].last_written_ts;
         // if the packet doesn't have a dts, use the pts. if it doesn't have a pts, consider it to have the same timestamp as the last packet
         let ts = p.dts_90khz.or(p.pts_90khz);
@@ -169,7 +169,7 @@ mod test {
                         .write(
                             0,
                             Packet {
-                                data: frame.data.into(),
+                                data: &frame.data,
                                 random_access_indicator,
                                 pts_90khz: frame.header.optional_header.as_ref().and_then(|h| h.pts),
                                 dts_90khz: frame.header.optional_header.as_ref().and_then(|h| h.dts),
diff --git a/mpeg2/src/muxer.rs b/mpeg2/src/muxer.rs
index c2f14a3..96fcafd 100644
--- a/mpeg2/src/muxer.rs
+++ b/mpeg2/src/muxer.rs
@@ -190,8 +190,8 @@ impl<W: Write> Muxer<W> {
 }

 #[derive(Default)]
-pub struct Packet {
-    pub data: Vec<u8>,
+pub struct Packet<'a> {
+    pub data: &'a [u8],
     pub random_access_indicator: bool,
     pub pts_90khz: Option<u64>,
     pub dts_90khz: Option<u64>,
@@ -247,7 +247,7 @@ mod test {
                         .write(
                             &mut video_out,
                             Packet {
-                                data: frame.data.into(),
+                                data: &frame.data,
                                 random_access_indicator,
                                 pts_90khz: frame.header.optional_header.as_ref().and_then(|h| h.pts),
                                 dts_90khz: frame.header.optional_header.as_ref().and_then(|h| h.dts),

```

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,189 @@
...
+            largest_ts_in_buffer: 0,
+            streams: vec![],
+        }
+    }
+
+    pub fn add_stream(&mut self, config: StreamConfig) {
+        let wrapper = self.muxer.new_stream(config);
+        let interleaving_stream = InterleavingStream {
+            wrapper,
+            buffered_packets: VecDeque::with_capacity(6),
```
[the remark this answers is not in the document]
lzuosym: doesn't have to be `6`. would like to pre-allocate some memory for the buffer. Not sure if it to reduce the number of vector reallocations, maybe I can just use `VecDeque::new()`

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,189 @@
...
+
+pub struct InterleavingMuxer<W> {
+    muxer: Muxer<W>,
+    max_buffer_duration: u64,
+    largest_ts_in_buffer: u64,
+    streams: Vec<InterleavingStream>,
+}
+
+pub struct InterleavingStream {
+    wrapper: Stream,
```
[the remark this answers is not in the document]
lzuosym: `inner` is much clear 👍

mpeg2/src/interleaving_muxer.rs
lzuosym: 👍

mpeg2/src/interleaving_muxer.rs
lzuosym: seeing the following errors when calling `self.write_packet` directly, I couldn't find a way to walk around.
```bash
error[E0499]: cannot borrow `*self` as mutable more than once at a time
   --> mpeg2/src/interleaving_muxer.rs:114:21
    |
109 |         for (stream_index, stream) in self.streams.iter_mut().enumerate() {
    |                                       -----------------------------------
    |                                       |
    |                                       first mutable borrow occurs here
    |                                       first borrow later used here
...
114 |                     self.write_packet(stream_index, stream.buffered_packets.pop_front().unwrap());
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ second mutable borrow occurs here
```

mpeg2/src/interleaving_muxer.rs
lzuosym: I had this in mind in the beginning, but totally forget. Thanks for bringing it up. 👍

mpeg2/src/interleaving_muxer.rs
lzuosym: We definitely need tests for all these use cases. But first thing we need some out-of-sync streams for testing. I am not sure how to create them.

ccbrown: You can just create tests using fake data. It doesn't have to be actual audio and video. I would just create packets with timestamps that demonstrate all the scenarios we want to cover, and fill them with arbitrary data.

lzuosym: added some tests. but still thinking how to handle the timestamp roll over case.

mpeg2/src/interleaving_muxer.rs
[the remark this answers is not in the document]
ccbrown: In the absence of benchmark data, I would suggest just going with `VecDeque::new()`. Using 6 makes the reader wonder. And it seems unlikely there's a significant performance difference.

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,463 @@
...
+        self.streams.push(interleaving_stream);
+    }
+
+    #[inline]
+    fn get_most_significant_bit(&self, mut n: u64) -> u64 {
+        while n > 1 && (n & (n - 1) != 0) {
+            n &= n - 1;
+        }
+        n
+    }
```
ccbrown, line 51: No need for `&self` here. And there are standard library functions that can be used here to avoid the loop and use intrinsics instead:

```suggestion
    #[inline]
    fn get_most_significant_bit(n: u64) -> u64 {
        (1 << (64 - n.leading_zeros())) >> 1
    }
```

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    fn handle_ts_rollover(&mut self, stream_index: usize, packet: &mut Packet) {
+        let s = &self.streams[stream_index];
+        let buffered_packets = &s.buffered_packets;
+        if let Some(ref mut dts) = packet.dts_90khz {
+            let largest_ts = buffered_packets
+                .back()
+                .map_or(s.last_written_ts, |p| p.dts_90khz.or(p.pts_90khz).unwrap_or(s.last_written_ts));
```
ccbrown, line 59: I think this can be simplified:

```suggestion
                .and_then(|p| p.dts_90khz.or(p.pts_90khz)).unwrap_or(s.last_written_ts);
```

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,463 @@
...
+            if *dts < largest_ts && largest_ts - *dts > TS_ROLLOVER_THREAD {
+                *dts |= self.get_most_significant_bit(largest_ts | TS_33BIT_MASK) << 1;
+            }
+        } else if let Some(ref mut pts) = packet.pts_90khz {
+            let largest_ts = buffered_packets
+                .back()
+                .map_or(s.last_written_ts, |p| p.dts_90khz.or(p.pts_90khz).unwrap_or(s.last_written_ts));
```
ccbrown, line 66: Same.

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,463 @@
...
+                _ => {}
+            }
+        }
+        Ok(())
+    }
+}
+
+impl<W: Write> Drop for InterleavingMuxer<W> {
+    fn drop(&mut self) {
+        self.flush(true).expect("Errors are ignored when flushing the buffered packets");
```
ccbrown, line 209: I would recommend getting rid of the `ignore_error` argument and just ignoring the errors like this:

```suggestion
        // Errors are ignored when flushing the buffered packets.
        let _ = self.flush();
```

It makes `flush` simpler, and if `flush` encounters an error, it should return immediately instead of continuing to try to write the other packets.

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,463 @@
...
+
+            muxer.flush(false).unwrap();
+            assert!(muxer.streams.iter().all(|s| s.buffered_packets.is_empty()));
+            assert!(muxer.streams.iter().all(|s| s.last_written_ts == 0));
+            assert_eq!(muxer.largest_ts_in_buffer, 0);
+        }
+        let reconstructed_pes_packets =
+            get_pes_packets_from_byte_stream(&data_out, num_of_pes_packets_per_stream * num_of_streams, &[0x100, 0x101, 0x102], false);
+
+        //
```
ccbrown, line 340:
```suggestion
```

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,463 @@
...
+                .map_or(s.last_written_ts, |p| p.dts_90khz.or(p.pts_90khz).unwrap_or(s.last_written_ts));
+            if *dts < largest_ts && largest_ts - *dts > TS_ROLLOVER_THREAD {
+                *dts |= self.get_most_significant_bit(largest_ts | TS_33BIT_MASK) << 1;
+            }
+        } else if let Some(ref mut pts) = packet.pts_90khz {
+            let largest_ts = buffered_packets
+                .back()
+                .map_or(s.last_written_ts, |p| p.dts_90khz.or(p.pts_90khz).unwrap_or(s.last_written_ts));
+            if *pts < largest_ts && largest_ts - *pts > TS_ROLLOVER_THREAD {
+                *pts |= self.get_most_significant_bit(largest_ts | TS_33BIT_MASK) << 1;
```
ccbrown, line 68: I don't think this operation is quite right. I tried adding a print to see what the resulting timestamps were and they look like they're growing out of control:

```diff
diff --git a/mpeg2/src/interleaving_muxer.rs b/mpeg2/src/interleaving_muxer.rs
index 95f995d..cfa9d0a 100644
--- a/mpeg2/src/interleaving_muxer.rs
+++ b/mpeg2/src/interleaving_muxer.rs
@@ -72,6 +72,7 @@ impl<W: Write> InterleavingMuxer<W> {
 
     pub fn write(&mut self, stream_index: usize, mut p: Packet) -> Result<(), EncodeError> {
         self.handle_ts_rollover(stream_index, &mut p);
+        println!("fixed dts: {:?} pts: {:?}", p.dts_90khz, p.pts_90khz);
 
         let last_written_ts = self.streams[stream_index].last_written_ts;
         // if the packet doesn't have a dts, use the pts. if it doesn't have a pts, consider it to have the same timestamp as the last packet
```

```
chris@Chriss-MBP av-rs % cargo test -p mpeg2 -- test_timestamp_rollover_in_multiple_streams --nocapture
   Compiling mpeg2 v0.1.0 (/Users/chris/repos/github.com/sportsball-ai/av-rs/mpeg2)
    Finished test [unoptimized + debuginfo] target(s) in 0.33s
     Running unittests (target/debug/deps/mpeg2-69d0e94d8acca5ae)

running 1 test
fixed dts: Some(2254684) pts: Some(165039)
fixed dts: Some(2253182) pts: Some(157532)
fixed dts: None pts: Some(2257687)
fixed dts: Some(2256185) pts: Some(162036)
fixed dts: Some(2260690) pts: Some(171045)
fixed dts: Some(2259188) pts: Some(163538)
fixed dts: None pts: Some(2263693)
fixed dts: Some(2262191) pts: Some(168042)
fixed dts: Some(2266696) pts: Some(177051)
fixed dts: Some(2265194) pts: Some(169544)
fixed dts: None pts: Some(2269699)
fixed dts: Some(2268197) pts: Some(174048)
fixed dts: Some(2272702) pts: Some(183057)
fixed dts: Some(2271200) pts: Some(175550)
fixed dts: None pts: Some(2275705)
fixed dts: Some(2274203) pts: Some(180054)
fixed dts: Some(2278708) pts: Some(189063)
fixed dts: Some(2277206) pts: Some(181556)
fixed dts: None pts: Some(2281711)
fixed dts: Some(2280209) pts: Some(186060)
fixed dts: Some(8590062093) pts: Some(135009)
fixed dts: Some(8590060592) pts: Some(129003)
fixed dts: None pts: Some(17179999689)
fixed dts: Some(17179998187) pts: Some(132006)
fixed dts: Some(34359871876) pts: Some(141015)
fixed dts: Some(34359870374) pts: Some(133508)
fixed dts: None pts: Some(68719613247)
fixed dts: Some(68719611745) pts: Some(138012)
fixed dts: Some(137439092986) pts: Some(147021)
fixed dts: Some(137439091484) pts: Some(139514)
fixed dts: None pts: Some(274878049461)
fixed dts: Some(274878047959) pts: Some(144018)
fixed dts: Some(549755959408) pts: Some(153027)
fixed dts: Some(549755957906) pts: Some(145520)
fixed dts: None pts: Some(1099511776299)
fixed dts: Some(1099511774797) pts: Some(150024)
fixed dts: Some(2199023407078) pts: Some(159033)
fixed dts: Some(2199023405576) pts: Some(151526)
fixed dts: None pts: Some(4398046665633)
fixed dts: Some(4398046664131) pts: Some(156030)
test interleaving_muxer::test::test_timestamp_rollover_in_multiple_streams ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 7 filtered out; finished in 0.02s
```

The timestamps quickly grow to 0x40000025BA1, and if we kept going, I'd guess the muxer would break entirely.

ccbrown (changes requested):

mpeg2/src/interleaving_muxer.rs
lzuosym: 👍  I didn't know there is such a handy method in rust!

mpeg2/src/interleaving_muxer.rs
lzuosym: 👍

mpeg2/src/interleaving_muxer.rs
lzuosym: Agree!

mpeg2/src/interleaving_muxer.rs
lzuosym: Good catch! Thank you! Should have used more packets in test to catch this bug! Timestamp rollover is a little bit tricky. Need to think it over carefully.

mpeg2/src/interleaving_muxer.rs
```diff
@@ -0,0 +1,479 @@
...
+    fn handle_ts_rollover(&mut self, stream_index: usize, packet: &mut Packet) {
+        let stream = &mut self.streams[stream_index];
+        if stream.ts_rollover {
+            if let Some(ref mut dts) = packet.dts_90khz {
+                *dts |= TS_ROLLOVER_FLAG;
+            } else if let Some(ref mut pts) = packet.pts_90khz {
+                *pts |= TS_ROLLOVER_FLAG;
+            }
+            return;
+        }
```
ccbrown, line 57: This will only work for a single rollover right? If the stream runs for multiple days, there could be multiple rollovers. Could we add another test case for that?

ccbrown (commented):
I believe the logic for overflows should look something like this:

```rust
/// This function returns a timestamp with high bits added in order to keep it as close as possible to `last_fixed_timestamp`. To correctly track overflows, `last_fixed_timestamp` should be the timestamp that was previously returned by this function.
fn fixed_timestamp(ts: u64, last_fixed_timestamp: u64) u64 {
    if (last_fixed_timestamp & TS_33BIT_MASK) > 0x180000000 && ts < 0x80000000 {
        // It looks like ts rolled over. This must be the first timestamp of a new epoch.
        ts | ((last_fixed_timestamp >> 33) + 1) << 33
    } else if (last_fixed_timestamp & TS_33BIT_MASK) < 0x80000000 && ts > 0x180000000 && last_fixed_timestamp > TS_33BIT_MASK {
        // It looks like ts rolled under. This could happen if we look at timestamps out of order.
        ts | ((last_fixed_timestamp >> 33) - 1) << 33
    } else {
        ts
    }
}
```

Basically we want to keep the timestamps as smooth and continuous as possible by adding the high bits necessary to minimize the deltas between the timestamps we output.

lzuosym:
> I believe the logic for overflows should look something like this:
> 
> ```rust
> /// This function returns a timestamp with high bits added in order to keep it as close as possible to `last_fixed_timestamp`. To correctly track overflows, `last_fixed_timestamp` should be the timestamp that was previously returned by this function.
> fn fixed_timestamp(ts: u64, last_fixed_timestamp: u64) u64 {
>     if (last_fixed_timestamp & TS_33BIT_MASK) > 0x180000000 && ts < 0x80000000 {
>         // It looks like ts rolled over. This must be the first timestamp of a new epoch.
>         ts | ((last_fixed_timestamp >> 33) + 1) << 33
>     } else if (last_fixed_timestamp & TS_33BIT_MASK) < 0x80000000 && ts > 0x180000000 && last_fixed_timestamp > TS_33BIT_MASK {
>         // It looks like ts rolled under. This could happen if we look at timestamps out of order.
>         ts | ((last_fixed_timestamp >> 33) - 1) << 33
>     } else {
>         ts
>     }
> }
> ```
> 
> Basically we want to keep the timestamps as smooth and continuous as possible by adding the high bits necessary to minimize the deltas between the timestamps we output.

Is `last_fixed_timestamp` for the timestamps in all the streams? How do we update the `last_fixed_timestamp`? if it's a non-decreasing variable, after some time, the first part in the `if` condition `if (last_fixed_timestamp & TS_33BIT_MASK) > 0x180000000 && ts < 0x80000000 ` will be always true.  And the `else if....` condition will always be false.

ccbrown:
> Is last_fixed_timestamp for the timestamps in all the streams? How do we update the last_fixed_timestamp?

It could be. The only thing that's important here is that it's a _recent_ timestamp. Then the output of the function will never have any big jumps due to overflows.

> if it's a non-decreasing variable, after some time, the first part in the if condition if (last_fixed_timestamp & TS_33BIT_MASK) > 0x180000000 && ts < 0x80000000 will be always true. And the else if.... condition will always be false.

I don't believe that's true. For example:

Say `last_fixed_timestamp` = 0x5000 and `ts` = 0x6000. `(last_fixed_timestamp & TS_33BIT_MASK) > 0x180000000 && ts < 0x80000000` would be false.

Say `last_fixed_timestamp` = 0x1ffffffffffffffff and `ts` = 0x1000. `(last_fixed_timestamp & TS_33BIT_MASK) > 0x180000000 && ts < 0x80000000` would be true, and the output would be `0x200000000000001000`.

Say `last_fixed_timestamp` = 0x200000000000005000 and `ts` = 0x6000. `(last_fixed_timestamp & TS_33BIT_MASK) > 0x180000000 && ts < 0x80000000` would be false. Now I'm realizing there is a bug here, because the output should be 0x200000000000006000. The condition is correct, but the last case should return `ts | (last_fixed_timestamp >> 33) << 33`.

```rust
/// This function returns a timestamp with high bits added in order to keep it as close as possible to `last_fixed_timestamp`. To correctly track overflows, `last_fixed_timestamp` should be the timestamp that was previously returned by this function.
fn fixed_timestamp(ts: u64, last_fixed_timestamp: u64) u64 {
    if (last_fixed_timestamp & TS_33BIT_MASK) > 0x180000000 && ts < 0x80000000 {
        // It looks like ts rolled over. This must be the first timestamp of a new epoch.
        ts | ((last_fixed_timestamp >> 33) + 1) << 33
    } else if (last_fixed_timestamp & TS_33BIT_MASK) < 0x80000000 && ts > 0x180000000 && last_fixed_timestamp > TS_33BIT_MASK {
        // It looks like ts rolled under. This could happen if we look at timestamps out of order.
        ts | ((last_fixed_timestamp >> 33) - 1) << 33
    } else {
        ts | (last_fixed_timestamp >> 33) << 33
    }
}
```

mpeg2/src/muxer.rs
```diff
@@ -156,64 +172,41 @@ impl<W: Write> Muxer<W> {
...
-        p.encode(&mut state.w)?;
+        p.encode(&mut self.w)?;
         Ok(())
     }
 }
 
 #[derive(Default)]
-pub struct Packet<'a> {
-    pub data: &'a [u8],
+pub struct Packet {
```
ccbrown, line 193: I think everything here looks good, but I want to revisit this. What would you think about using `Cow` here?

```diff
diff --git a/mpeg2/src/interleaving_muxer.rs b/mpeg2/src/interleaving_muxer.rs
index c8c5a24..9593eb4 100644
--- a/mpeg2/src/interleaving_muxer.rs
+++ b/mpeg2/src/interleaving_muxer.rs
@@ -17,7 +17,7 @@ pub struct InterleavingMuxer<W: Write> {
 
 pub struct InterleavingStream {
     inner: Stream,
-    buffered_packets: VecDeque<Packet>,
+    buffered_packets: VecDeque<Packet<'static>>,
     last_written_ts: u64,
 }
 
@@ -76,7 +76,7 @@ impl<W: Write> InterleavingMuxer<W> {
             }
         } else {
             // and then we need to see if this enabled any other streams to emit their outputs
-            self.streams[stream_index].buffered_packets.push_back(p);
+            self.streams[stream_index].buffered_packets.push_back(p.into_owned());
             if let Some(t) = ts {
                 self.largest_ts_in_buffer = self.largest_ts_in_buffer.max(t);
             }
diff --git a/mpeg2/src/muxer.rs b/mpeg2/src/muxer.rs
index c2f14a3..66994b3 100644
--- a/mpeg2/src/muxer.rs
+++ b/mpeg2/src/muxer.rs
@@ -1,5 +1,5 @@
 use super::{pes, ts, EncodeError};
-use alloc::vec::Vec;
+use alloc::{borrow::Cow, vec::Vec};
 use core2::io::Write;
 
 struct StreamState {
@@ -190,13 +190,24 @@ impl<W: Write> Muxer<W> {
 }
 
 #[derive(Default)]
-pub struct Packet {
-    pub data: Vec<u8>,
+pub struct Packet<'a> {
+    pub data: Cow<'a, [u8]>,
     pub random_access_indicator: bool,
     pub pts_90khz: Option<u64>,
     pub dts_90khz: Option<u64>,
 }
 
+impl<'a> Packet<'a> {
+    pub fn into_owned(self) -> Packet<'static> {
+        Packet {
+            data: Cow::Owned(self.data.into_owned()),
+            random_access_indicator: self.random_access_indicator,
+            pts_90khz: self.pts_90khz,
+            dts_90khz: self.dts_90khz,
+        }
+    }
+}
+
 pub struct Stream {
     packet_id: u16,
     continuity_counter: u8,
```

We have at least one place where this would avoid unnecessary copies because the data is owned by a NETINT buffer instead of a `Vec`.

lzuosym: I struggled fighting with rust compiler,  but failed to figure out how to keep the lifetime parameter in `Packet`. I think this static lifetime trick is useful! 👍

ccbrown (approved):

PR #86, opened 2022-05-10
AVS-211 add wall-clock timestamps to video streams

lzuosym:
Currently, the AV service timestamps video based on when that video is received, which makes the timestamps subject to network and other types of upstream jitter.

This PR makes it possible for av-agent to embed accurate wall-clock timestamps in TS packets.

For context, see the paper [Understanding Timelines within MPEG Standards](https://ieeexplore.ieee.org/document/7293587) and ISO/IEC 13818-1:2019(E) Annex U `Carriage of timeline and external media information (TEMI) over MPEG-2 transport streams`


mpeg2/src/temi.rs
```diff
@@ -0,0 +1,290 @@
...
+            buf[2] |= 0b0000_0001;
+        }
+        if self.discontinuity.unwrap_or(false) {
+            buf[3] = 0b1000_0000;
+        }
+        buf[4] = self.timeline_id;
+
+        let mut ret = 5;
+        if let Some(has_timestamp) = self.has_timestamp {
+            buf[5] = self.timescale as _;
```
dotcypress, line 74: I think we can this code more readable with something like `BigEndian::read_u32(&mut buf[5..8), self.timescale)`. And `BigEndian::read_u32(&buf)` to converting back.

lzuosym: I was supposed to add Edward to review PR. Sorry for getting you involved. Thanks for reviewing the code! Is `BigEndian` from `byteorder` crate?

dotcypress: Yes, BigEndian from byteorder crate.

lzuosym: 👍

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        if self.force_reload.unwrap_or(false) {
+            buf[2] |= 0b0000_0010;
+        }
+        if self.paused.unwrap_or(false) {
+            buf[2] |= 0b0000_0001;
+        }
+        if self.discontinuity.unwrap_or(false) {
+            buf[3] = 0b1000_0000;
+        }
```
ccbrown, line 70: Shouldn't this be...

```suggestion
        if self.force_reload.unwrap_or(false) {
            buf[2] |= 0b0010_0000;
        }
        if self.paused.unwrap_or(false) {
            buf[2] |= 0b0100_0000;
        }
        if self.discontinuity.unwrap_or(false) {
            buf[3] = 0b1000_0000;
        }
```

?

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,275 @@
...
+            } else if has_timestamp != 1 {
+                return Err(EncodeError::other("Invalid has_timestamp value"));
+            }
+        }
+
+        if let Some(ntp_ts) = self.ntp_timestamp {
+            buf[2] |= 0b0010_0000;
+            BigEndian::write_u64(&mut buf[ret..(ret + 8)], ntp_ts);
+            ret += 8;
+        } else if let Some(ptp_ts) = self.ptp_timestamp {
```
ccbrown, line 90: I don't see any reason we couldn't have both of these:

```suggestion
        }

        if let Some(ptp_ts) = self.ptp_timestamp {
```

mpeg2/src/temi.rs
lzuosym: image
I think `paused` is the lowest-order bit in `buf[2]`.

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,275 @@
...
+use byteorder::{BigEndian, ByteOrder};
+use core2::io::Write;
+
+const AF_DESCR_TAG_TIMELINE: u8 = 0x04;
+
+#[derive(Debug, Default, PartialEq)]
+pub struct TEMITimelineDescriptor {
+    pub has_timestamp: Option<u8>,
+    pub timescale: u32,
+    pub media_timestamp: u64,
```
ccbrown, line 11: It seems like you could get rid of this field. You could instead make `media_timestamp` an `Option<u64>`, and during encoding, say:

```rust
let has_timestamp = media_timestamp.map(|ts| if ts >> 32 != 0 { 2 } else { 1 }).unwrap_or(0);
```

mpeg2/src/temi.rs
lzuosym: should be able to have both! 👍

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    pub has_timecode: Option<u8>,
+    pub drop: bool,
+    pub frames_per_tc_seconds: u16,
+    pub duration: u16,
+    pub time_code: u64,
+
+    pub force_reload: Option<bool>,
+    pub paused: Option<bool>,
+    pub discontinuity: Option<bool>,
```
ccbrown, line 24: Why not just make these `bool`s?

```suggestion
    pub force_reload: bool,
    pub paused: bool,
    pub discontinuity: bool,
```

mpeg2/src/temi.rs
ccbrown: Ah, you're right. I read that wrong. 👍

mpeg2/src/temi.rs
lzuosym: I don't see any reason to use `bool` instead of `Option<bool>`.  Encoded `None` valued won't be decoded to `None`.    Will remove `Option`.

mpeg2/src/ts.rs
```diff
@@ -18,6 +19,7 @@ pub struct AdaptationField {
     pub discontinuity_indicator: Option<bool>,
     pub random_access_indicator: Option<bool>,
     pub program_clock_reference_27mhz: Option<u64>,
+    pub temi_timeline_descriptor: Option<TEMITimelineDescriptor>,
```
ccbrown, line 22: As far as I can tell, an adaptation field can have multiple timeline descriptors. Should this be a `Vec` instead of `Option`?

lzuosym: you are right, the standards support multiple timeline descriptors. But in reality do we need to embed more than 2 TEMITimelineDescriptors with different timestamps in one TS packet? I though we might need `Vec` for different types of TEMI descriptors. For simplicity purposes, only one timeline descriptor is in here.

ccbrown: We might not have any current use-cases for multiple timeline descriptors, but I'd like to keep this crate as unopinionated and as close to the spec as possible.

lzuosym: sure, shouldn't be a problem to make it `Vec`.

mpeg2/src/temi.rs
lzuosym: This seems to be wrong. When I decode a packet from a real video clip. `has_timestamp=2`, but `media_timestamp` is `0`.  It is the value of `has_timestamp` that determines the presence of `media_timestamp` and `timescale`.
image

ccbrown: Hmm... It seems like it would still be reasonable to omit `has_timestamp` from the struct. I can't imagine a user of the library ever wanting to use a 64 bit timestamp when 32 bits is enough or ever wanting to know whether a timestamp was encoded as 32 or 64 bits. But in the interest of being un-opinionated, I suppose leaving it is fine too.

I'll just make one ask here then: Either define some constants for `has_timestamp`'s possible values or make it an `enum` type (e.g. with `None`, `ThirtyTwoBits`, `SixtyFourBits`, and `Reserved` variants).

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,282 @@
+use crate::{DecodeError, EncodeError};
+use byteorder::{BigEndian, ByteOrder};
+use core2::io::Write;
+
+pub const AF_DESCR_TAG_TIMELINE: u8 = 0x04;
+
+#[derive(Debug, PartialEq, PartialOrd, Copy, Clone)]
+#[allow(non_camel_case_types)]
+#[repr(C)]
```
ccbrown, line 9: Are these necessary?

```suggestion
```

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub time_code: u64,
+
+    pub force_reload: bool,
+    pub paused: bool,
+    pub discontinuity: bool,
+    pub timeline_id: u8,
+}
+
+impl TEMITimelineDescriptor {
+    pub fn encode_len(&self) -> usize {
```
ccbrown, line 44: In `AdaptationField`, this function is named "encoded_len". We should do the same here:

```suggestion
    pub fn encoded_len(&self) -> usize {
```

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub timescale: u32,
+    pub media_timestamp: u64,
+
+    pub ntp_timestamp: Option<u64>,
+    pub ptp_timestamp: Option<u128>,
+
+    pub drop: bool,
+    pub frames_per_tc_seconds: u16,
+    pub duration: u16,
+    pub has_timecode: TimestampLength,
```
ccbrown, line 34: You should make a separate enum type for this, since the timecode is either 24 or 64 bits (not 32 or 64 like the timestamp).

```suggestion
    pub has_timecode: TimecodeLength,
```

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        if self.has_timecode != TimestampLength::None {
+            buf[2] |= (self.has_timecode as u8) << 2;
+            if self.drop {
+                buf[ret] = 0b1000_0000;
+            }
+            buf[ret] |= (self.frames_per_tc_seconds >> 8) as u8 & 0b0111_1111;
+            buf[ret + 1] = self.frames_per_tc_seconds as _;
+            buf[ret + 2] = (self.duration >> 8) as u8;
+            buf[ret + 3] = self.duration as _;
```
ccbrown, line 118:
```suggestion
            BigEndian::write_u16(&mut buf[ret..], self.frames_per_tc_seconds & 0x7fff);
            BigEndian::write_u16(&mut buf[ret + 2..], self.duration);
            if self.drop {
                buf[ret] |= 0b1000_0000;
            }
```

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,282 @@
...
+        if let Some(ntp_ts) = self.ntp_timestamp {
+            buf[2] |= 0b0010_0000;
+            BigEndian::write_u64(&mut buf[ret..(ret + 8)], ntp_ts);
+            ret += 8;
+        }
+        if let Some(ptp_ts) = self.ptp_timestamp {
+            buf[2] |= 0b0001_0000;
+            BigEndian::write_u16(&mut buf[ret..(ret + 2)], (ptp_ts >> 64) as u16);
+            ret += 2;
+            BigEndian::write_u64(&mut buf[ret..(ret + 8)], ptp_ts as u64);
```
ccbrown, line 106: These write ranges don't need to have an end specified. I would recommend just removing them:

```suggestion
            BigEndian::write_u64(&mut buf[ret..], ptp_ts as u64);
```

Makes the code easier to read and check.

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,282 @@
...
+            } else {
+                BigEndian::write_u64(&mut buf[ret..(ret + 8)], self.time_code);
+                ret += 8;
+            }
+        }
+
+        buf[1] = ret as u8 - 2;
+
+        w.write_all(&buf)?;
+        Ok(buf.len())
```
ccbrown, line 135: We have a variable named `ret`. We should either return it or name it something else like "offset".

```suggestion
        Ok(ret)
```

mpeg2/src/ts.rs
```diff
@@ -613,4 +682,118 @@ mod test {
...
+        ];
+        let packet = Packet::decode(&packet_data).unwrap();
+        let af = packet.adaptation_field.unwrap();
+        let temi_descriptors = af.temi_timeline_descriptors;
+        assert_eq!(temi_descriptors.len(), 1);
+        let decoded_temi = &temi_descriptors[0];
+        let temi = TEMITimelineDescriptor {
+            has_timestamp: TimestampLength::SixtyFourBits,
+            timescale: 1000000,
+            paused: true,
```
ccbrown, line 784: I think later in that test file, after about 10 seconds, the timeline unpauses, and you should start seeing actual timestamps. Can you check and see if you can find some TEMIs with actual, non-zero timestamps that we can add tests for?

mpeg2/src/ts.rs
```diff
@@ -36,21 +39,66 @@ impl AdaptationField {
...
+        let mut len = 1;
+        if self.discontinuity_indicator.is_some() || self.random_access_indicator.is_some() || self.program_clock_reference_27mhz.is_some() {
+            len += 1;
+        }
+        if self.program_clock_reference_27mhz.is_some() {
+            len += 6;
+        }
+        if !self.private_data_types.is_empty() {
+            len += 1 + self.private_data_types.len();
+        }
```
ccbrown, line 84: Don't we also need to add 2 bytes for `adaptation_field_extension_length` and the flags after it?

```suggestion
        }
        if !self.temi_timeline_descriptors.is_empty() {
                len += 2;
        }
```

mpeg2/src/ts.rs
```diff
@@ -36,21 +39,66 @@ impl AdaptationField {
             } else {
                 None
             };
+            let mut n = if af.program_clock_reference_27mhz.is_some() { 8 } else { 2 };
+            if buf[1] & 0x02 != 0 {
+                let transport_private_date_length = buf[n];
```
ccbrown, line 44: Typo:

```suggestion
                let transport_private_data_length = buf[n];
```

mpeg2/src/ts.rs
```diff
@@ -36,21 +39,66 @@ impl AdaptationField {
             } else {
                 None
             };
+            let mut n = if af.program_clock_reference_27mhz.is_some() { 8 } else { 2 };
```
ccbrown, line 42: We also need to skip additional bytes if the OPCR is present:

```suggestion
            if buf[1] & 0x08 != 0 {
                        n += 6; // skip over the OPCR
            }            
```

mpeg2/src/ts.rs
ccbrown, line 44: Can we add some length checks? Before we read from `buf[n]`, let's make sure buf is long enough. This function shouldn't panic on invalid input.

mpeg2/src/ts.rs
```diff
@@ -36,21 +39,66 @@ impl AdaptationField {
...
+                let transport_private_date_length = buf[n];
+                af.private_data_types = buf[n..(n + transport_private_date_length as usize)].to_vec();
+                n += transport_private_date_length as usize + 1
+            }
+            if buf[1] & 1 == 1 {
+                if buf[n] == 0 {
+                    return Err(DecodeError::new("invalid adaptation_field_extension_length"));
+                }
+                let temi_descriptors_len = (buf[n] - 1) as usize;
+                n += 2;
```
ccbrown, line 53: We need to check the LTW, piecewise rate, and seamless splice flags, and skip over additional bytes if any of those are non-zero.

ccbrown (changes requested):

mpeg2/src/temi.rs
lzuosym: No! they are not. will remove them.

mpeg2/src/temi.rs
lzuosym: 👍

mpeg2/src/temi.rs
lzuosym: `TimestampLength` enum is used by both `has_timecode` and `has_timestamp`, maybe rename this enum to `TimeLength`?

mpeg2/src/temi.rs
lzuosym: Ah. this is much cleaner! 👍

mpeg2/src/temi.rs
lzuosym: will remove all the redundant end indexes.

mpeg2/src/temi.rs
lzuosym: they're equal, should simplify return `ret`.

mpeg2/src/ts.rs
lzuosym: I don't see any `ntp_timestamp` or `ptp_timestamp` values set in those 2 test files.

mpeg2/src/ts.rs
lzuosym: okay. I think we also need to skip `splice_countdown` if its flag is 1

mpeg2/src/ts.rs
lzuosym: The length was missed. will add them.

mpeg2/src/temi.rs
ccbrown: My point is that it shouldn't be used for both of them. Because the timecode is either 24 bits or 64 bits, but the timestamp is either 32 or 64 bits. They don't have the same possible lengths.

lzuosym: I see you point, maybe we can define the `enum` as follows,
```rust
pub enum TimeFieldLength {
    None = 0,
    Short = 1,
    Long = 2,
}
```
Then it makes sense to be used by both of them. `Short` means length is less than or equal to 32 bits.

ccbrown: 🤷‍♂️ Either works.

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,279 @@
...
+            ts |= BigEndian::read_u64(&buf[n..]) as u128;
+            n += 8;
+            ret.ptp_timestamp = Some(ts);
+        }
+
+        match (buf[2] & 0b0000_1100) >> 2 {
+            0 => ret.has_timecode = TimeFieldLength::None,
+            1 => ret.has_timecode = TimeFieldLength::Short,
+            2 => ret.has_timecode = TimeFieldLength::Long,
+            _ => return Err(DecodeError::new("incorrect has_timecode")),
```
ccbrown, line 186: A value of 3 is reserved:

> The term "reserved", when used in the clauses defining the coded bit stream, indicates that the value may be used in the future for ISO defined extensions.

If we want to avoid breakage in the future, we should interpret things exactly how the spec says. Meaning, if this is 3, we should parse the drop, frames_per_tc_seconds, and duration fields, but not parse any timecode.

Same for timestamp.

mpeg2/src/ts.rs
```diff
@@ -27,30 +30,112 @@ impl AdaptationField {
...
+                    af.temi_timeline_descriptors = vec![];
+                    let temi_buf = &buf[n..(n + temi_descriptors_len as usize)];
+                    let mut m = 0;
+                    while m + 2 <= temi_buf.len() {
+                        if temi_buf[m] == AF_DESCR_TAG_TIMELINE {
+                            let descr = TEMITimelineDescriptor::decode(&temi_buf[m..])?;
+                            m += 2 + temi_buf[m + 1] as usize;
+                            af.temi_timeline_descriptors.push(descr);
+                        } else {
+                            // Other types of TEMI descriptors are ignored
```
ccbrown, line 98:
```suggestion
                            // Other types of AF descriptors are ignored
```

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,279 @@
...
+        buf[4] = self.timeline_id;
+
+        let mut ret = 5;
+        if self.has_timestamp != TimeFieldLength::None {
+            BigEndian::write_u32(&mut buf[5..9], self.timescale);
+            if self.has_timestamp == TimeFieldLength::Short {
+                BigEndian::write_u32(&mut buf[9..13], self.media_timestamp as u32);
+                ret += 8;
+            } else if self.has_timestamp == TimeFieldLength::Long {
+                BigEndian::write_u64(&mut buf[9..17], self.media_timestamp);
```
ccbrown, line 90:
```suggestion
            BigEndian::write_u32(&mut buf[5..], self.timescale);
            if self.has_timestamp == TimeFieldLength::Short {
                BigEndian::write_u32(&mut buf[9..], self.media_timestamp as u32);
                ret += 8;
            } else if self.has_timestamp == TimeFieldLength::Long {
                BigEndian::write_u64(&mut buf[9..], self.media_timestamp);
```

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        let mut n = 5;
+        if ret.has_timestamp != TimeFieldLength::None {
+            ret.timescale = BigEndian::read_u32(&buf[5..9]);
+            if ret.has_timestamp == TimeFieldLength::Short {
+                ret.media_timestamp = BigEndian::read_u32(&buf[9..=12]) as u64;
+                n += 8;
+            } else {
+                ret.media_timestamp = BigEndian::read_u64(&buf[9..17]);
```
ccbrown, line 164:
```suggestion
            ret.timescale = BigEndian::read_u32(&buf[5..]);
            if ret.has_timestamp == TimeFieldLength::Short {
                ret.media_timestamp = BigEndian::read_u32(&buf[9..]) as u64;
                n += 8;
            } else {
                ret.media_timestamp = BigEndian::read_u64(&buf[9..]);
```

mpeg2/src/ts.rs
```diff
@@ -27,30 +30,112 @@ impl AdaptationField {
...
+                // skip splice_countdown
+                n += 1;
+            }
+            if buf[1] & 0b0000_0010 != 0 {
+                if n >= buf.len() || buf[n] as usize + n + 1 > buf.len() {
+                    return Err(DecodeError::new("adaptation field length too long"));
+                }
+                let transport_private_data_length = buf[n];
+                af.private_data_types = buf[n..(n + transport_private_data_length as usize)].to_vec();
+                n += transport_private_data_length as usize + 1
```
ccbrown, line 59:
```suggestion
                if n >= buf.len() || buf[n] as usize + n + 1 > buf.len() {
                    return Err(DecodeError::new("transport private data length too long"));
                }
                let transport_private_data_length = buf[n] as usize;
                af.private_data_types = buf[n..(n + transport_private_data_length)].to_vec();
                n += transport_private_data_length + 1
```

mpeg2/src/ts.rs
```diff
@@ -27,30 +30,112 @@ impl AdaptationField {
...
+            }
+            if buf[1] & 1 == 1 {
+                if n + 2 > buf.len() {
+                    return Err(DecodeError::new("adaptation field length too long"));
```
ccbrown, line 63:
```suggestion
                    return Err(DecodeError::new("adaptation field too short for adaptation field extension"));
```

I want to make sure that if we see one of these errors, we can find out where it actually came from. Ideally no two error messages are the same.

mpeg2/src/ts.rs
```diff
@@ -27,30 +30,112 @@ impl AdaptationField {
...
+            if buf[1] & 0b0000_0010 != 0 {
+                if n >= buf.len() || buf[n] as usize + n + 1 > buf.len() {
+                    return Err(DecodeError::new("adaptation field length too long"));
+                }
+                let transport_private_data_length = buf[n];
+                af.private_data_types = buf[n..(n + transport_private_data_length as usize)].to_vec();
+                n += transport_private_data_length as usize + 1
+            }
+            if buf[1] & 1 == 1 {
+                if n + 2 > buf.len() {
```
ccbrown, line 62: > The adaptation_field_length is an 8-bit field specifying the number of bytes in the adaptation_field immediately following the adaptation_field_length. The value '0' is for inserting a single stuffing byte in the adaptation field of a transport stream packet.

It looks like it's valid for the adaptation field extension to be only 1 byte.

mpeg2/src/ts.rs
```diff
@@ -27,30 +30,112 @@ impl AdaptationField {
...
+                    return Err(DecodeError::new("adaptation field length too long"));
+                } else if buf[n] == 0 {
+                    return Err(DecodeError::new("invalid adaptation_field_extension_length"));
+                }
+                let temi_descriptors_len = buf[n] - 1;
```
ccbrown, line 67: `buf[n]` is the `adaptation_field_extension_length` field, which includes the bytes required by the fields we skip below.

Let's change this variable so that it's clear that this is the length of the adaptation field extension and not just those descriptors:

```suggestion
                let adaptation_field_extension_length = buf[n];
```

Then fix the size of the buffer below.

ccbrown (changes requested):
I'm seeing a few more issues. Btw when the structures get as complex as this, I find it really helps to try to make the code look like the spec as much as possible. We've done a really good job of this in the h264 crate for example: https://github.com/sportsball-ai/av-rs/blob/main/h264/src/sequence_parameter_set.rs

It might be worth using some sort of bitstream reader like we do there so you can just write code like:

```
let adaptation_field_length = bitstream.read_u8()?;
if (adaptation_field_length > 0) {
    let discontinuity_indicator = bitstream.read_bit()?;
    let random_access_indicator = bitstream.read_bit()?;
    let elementary_stream_priority_indicator = bitstream.read_bit()?;
    let pcr_flag = bitstream.read_bit()?;
    let opcr_flag = bitstream.read_bit()?;

    ...

    if opcr_flag {
        bitstream.skip_bytes(6)?;
    }    
    if splicing_point_flag {
        bitstream.skip_bytes(1)?;
    }    
}
```

Then we can get rid of a large amount of the math and be able to verify correctness at a glance.

mpeg2/src/temi.rs
lzuosym: still missed a lot  :disappointed:

mpeg2/src/ts.rs
lzuosym: Absolutely! 👍

mpeg2/src/ts.rs
lzuosym: Interesting!  I don't see the single stuffing byte when `adaptation_field_length` is '0' in the pseudo code implementation. 
image

So for decode, if the `adaptation_field_length` is 0, we need to skip 1 byte in the `buf`
```
    pub fn decode(buf: &[u8]) -> Result<Self, DecodeError> {
        let mut af = Self::default();
        let af_length = buf[0];
        if af_length as usize > buf.len() - 1 {
            return Err(DecodeError::new("adaptation field length too long"));
        } else if buf[0] > 0 {
```

mpeg2/src/ts.rs
lzuosym: > I'm seeing a few more issues. Btw when the structures get as complex as this, I find it really helps to try to make the code look like the spec as much as possible. We've done a really good job of this in the h264 crate for example: https://github.com/sportsball-ai/av-rs/blob/main/h264/src/sequence_parameter_set.rs
> 
> It might be worth using some sort of bitstream reader like we do there so you can just write code like:
> 
> ```
> let adaptation_field_length = bitstream.read_u8()?;
> if (adaptation_field_length > 0) {
>     let discontinuity_indicator = bitstream.read_bit()?;
>     let random_access_indicator = bitstream.read_bit()?;
>     let elementary_stream_priority_indicator = bitstream.read_bit()?;
>     let pcr_flag = bitstream.read_bit()?;
>     let opcr_flag = bitstream.read_bit()?;
> 
>     ...
> 
>     if opcr_flag {
>         bitstream.skip_bytes(6)?;
>     }    
>     if splicing_point_flag {
>         bitstream.skip_bytes(1)?;
>     }    
> }
> ```
> 
> Then we can get rid of a large amount of the math and be able to verify correctness at a glance.

Hmm, this looks much nicer! I'll talk a look at h264 create, and try to make semi crate similar.

mpeg2/src/ts.rs
ccbrown: You know what... I read this wrong. I thought I was looking at the spec for `adaptation_field_extension_length`. I don't think anything needs to change here.

mpeg2/src/ts.rs
```diff
@@ -18,85 +20,205 @@ pub struct AdaptationField {
...
+                    bs.skip_bytes(5);
+                }
+
+                if !af_descriptor_not_present_flag {
+                    // AF descriptors
+                    if bs.remaining_bytes() + 1 < adaptation_field_extension_length as usize {
+                        return Err(DecodeError::new("adaptation descriptor field length too long"));
+                    }
+                    af.temi_timeline_descriptors = vec![];
+                    let mut af_descr_bitstream = bs.wrapped_stream(adaptation_field_extension_length as usize - 1);
```
ccbrown, line 106: `adaptation_field_extension_length` is the length of the adaptation field, including the LTW, piecewise rate, and seamless splice fields. So this wouldn't be right if those were present.

I think you should just shadow `bs` as soon as you read `adaptation_field_extension_length`:

```rust
let adaptation_field_extension_length = bs.read_u8();
let bs = bs.wrapped_stream(adaptation_field_extension_length as usize);
```

mpeg2/src/ts.rs
```diff
@@ -18,85 +20,205 @@ pub struct AdaptationField {
...
             return Err(DecodeError::new("adaptation field length too long"));
-        } else if buf[0] > 0 {
-            af.discontinuity_indicator = Some(buf[1] & 0x80 != 0);
-            af.random_access_indicator = Some(buf[1] & 0x40 != 0);
-            af.program_clock_reference_27mhz = if af_length >= 7 && buf[1] & 0x10 != 0 {
-                let base = (buf[2] as u64) << 25 | (buf[3] as u64) << 17 | (buf[4] as u64) << 9 | (buf[5] as u64) << 1 | (buf[6] as u64) >> 7;
-                let ext = ((buf[6] as u64) & 1) << 8 | (buf[7] as u64);
+        } else if af_length > 0 {
+            af.discontinuity_indicator = Some(bs.read_boolean());
+            af.random_access_indicator = Some(bs.read_boolean());
```
ccbrown, line 36: What would you think about making the `read_` methods return `Result` types so they could bubble up an error if there aren't enough bytes remaining? We could give them a `name` parameter so they could still provide meaningful errors:

```suggestion
            af.discontinuity_indicator = Some(bs.read_boolean("discontinuity_indicator")?);
            af.random_access_indicator = Some(bs.read_boolean("random_access_indicator")?);
```

This would allow us to get rid of the length checks from this function entirely and be a much more foolproof way to make sure it never panics.

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,271 @@
...
+            7
+        } else if self.has_timecode == TimeFieldLength::Long {
+            12
+        } else {
+            0
+        };
+        len
+    }
+
+    pub fn encode(&self, bs: &mut BitstreamWriter) -> Result<usize, EncodeError> {
```
ccbrown, line 68: It might be kind of nice to make `Bitstream` an implementation detail instead of a public API:

```suggestion
    pub fn encode(&self, buf: &[u8]) -> Result<usize, EncodeError> {
        let bs = Bitstream::new(buf);
```

ccbrown (changes requested):

mpeg2/src/ts.rs
lzuosym: Ahh, great catch! 👍

mpeg2/src/ts.rs
lzuosym: I haven't got used to idiomatic way to handle errors in Rust. will update the PR.

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,272 @@
...
+            paused,
+            discontinuity,
+            timeline_id,
+            ..Default::default()
+        };
+
+        if bs.remaining_bytes() + 2 + TEMI_TIMELINE_FLAGS_LENGTH < ret.encoded_len() {
+            return Err(DecodeError::new("not enough bytes to decode temi_timeline_descriptor"));
+        }
+
```
ccbrown, line 158: I'm not sure if this is correct, but it's not necessary anymore right?

```suggestion
```

mpeg2/src/temi.rs
```diff
@@ -0,0 +1,272 @@
...
+        Ok(len)
+    }
+}
+
+impl Decode for TEMITimelineDescriptor {
+    fn decode(bs: &mut Bitstream) -> Result<Self, DecodeError> {
+        if bs.remaining_bytes() < TEMI_TIMELINE_FLAGS_LENGTH {
+            return Err(DecodeError::new("not enough bytes for temi_timeline_descriptor flags"));
+        }
+
```
ccbrown, line 124: No need for this now, right?

```suggestion
```

mpeg2/src/ts.rs
```diff
@@ -18,85 +20,188 @@ pub struct AdaptationField {
...
+                // skip OPCR
+                bs.skip_bytes(6);
+            }
+            if splicing_point_flag {
+                // skip splice_countdown
+                bs.skip_bytes(1);
+            }
+            if transport_private_data_flag {
+                let transport_private_data_length = bs.read_u8("transport_private_data_length")?;
+                af.private_data_types = bs.read_n_bytes(transport_private_data_length as usize, "private_data_types")?.to_vec();
```
ccbrown, line 62: Typo maybe? Was this meant to be "bytes"?

```suggestion
                af.private_data_bytes = bs.read_n_bytes(transport_private_data_length as usize, "private_data_bytes")?.to_vec();
```

ccbrown (changes requested):
This is soo easy to review now! A few things.

mpeg2/src/temi.rs
lzuosym: yep, we don't need this check.

mpeg2/src/temi.rs
lzuosym: same

mpeg2/src/ts.rs
lzuosym: it's a typo!

ccbrown (approved):

PR #88, opened 2022-05-16
width / height calculation fix for h265 cropping windows

ccbrown:
We weren't using the right values for width/height when cropping windows were used. Before this change, this test file was interpreted as being 1920x1088 instead of 1920x1080.


lzuosym:
LGTM as long as there won't be overflow inside methods `croppedWidth`  and `croppedHeight`

PR #91, opened 2022-05-18
add failing test and fix for multi-stream interleaving latency

ccbrown:


mpeg2/src/interleaving_muxer.rs
```diff
@@ -435,4 +506,69 @@ mod test {
...
+        // packet after it.
+        muxer.write(1, simple_packet(50)).unwrap();
+        assert_eq!(w.packets().len(), 1);
+
+        // This packet still should not be written yet.
+        muxer.write(0, simple_packet(200)).unwrap();
+        assert_eq!(w.packets().len(), 1);
+
+        // At this point, we can write both the first packet and this one.
+        muxer.write(1, simple_packet(150)).unwrap();
```
lzuosym, line 552: At this time, there are 2 packets(ts=100, 200) in buffer, which belongs to stream(id=0). When adding a packet(ts=150, stream_id=1), it will be added to the buffer first and then the interleaving muxer checks the buffered packets in OTHER streams, and emits the packet if necessary. As a result, two packets(ts=50, 100) are written by muxer, and the other two(ts=150, 200) are still in buffer, one in each stream.

ccbrown: Yeah, that logic should be changed. There's no reason to buffer the packet with ts=150 at all in this scenario.

lzuosym: Okay, will submit a PR to fix it.

ccbrown: You can just commit to this PR. No need to make a separate one.

mpeg2/src/interleaving_muxer.rs
```diff
@@ -80,47 +80,7 @@ impl<W: Write> InterleavingMuxer<W> {
             if let Some(t) = ts {
```
ccbrown, line 80: @lzuosym Is there any way we can avoid the copy on line 79 in this scenario? If a packet can be emitted immediately, we shouldn't call `into_owned`, which is potentially a bit expensive.

lzuosym: After some thoughts, I think once if the packet is falling into the `else` condition, it will be kept in the buffer, and `emit_packets` method won't be able to emit it by all means  in the current round.

lzuosym: > After some thoughts, I think once if the packet is falling into the `else` condition, it will be kept in the buffer, and `emit_packets` method won't be able to emit it by all means in the current round.

There is one exception, when the packet from the stream, which doesn't have any packets buffered, it's possible that the new packet would be emitted immediately without being put into buffer.

mpeg2/src/interleaving_muxer.rs
```diff
@@ -174,14 +193,14 @@ impl<W: Write> InterleavingMuxer<W> {
...
             }
             stream.last_written_ts = 0;
         });
 
-        packets.sort_unstable_by_key(|(_, ts, p)| p.dts_90khz.or(p.pts_90khz).unwrap_or(*ts));
-        for (stream_index, _, packet) in packets {
-            self.write_muxer_packet(stream_index, packet)?;
+        packets.sort_unstable_by_key(|(_, p)| p.0);
+        for (stream_index, packet) in packets {
+            self.write_muxer_packet(stream_index, packet.2.expect("should not fail"))?;
```
ccbrown, line 203: The idea behind using `expect` over `unwrap` is that it encourages you to explain to the reader why it's okay to assume the value is `Some`. In this case, we can make the message a little more helpful:

```suggestion
            let packet = packet.2.expect("emit_packets should never leave None packets in the buffer");
            self.write_muxer_packet(stream_index, packet)?;
```

ccbrown:
@lzuosym This looks good to me. If you think it's ready, you can go ahead and give it an approval and merge it.

lzuosym:
:tada::tada::tada:

lzuosym:
oops. I clicked the merge button.

ccbrown:
Hmm... that shouldn't be possible. Looks like we never enabled branch protection on this repo.

lzuosym:
yeah, I was wondering how come I am able to merge my own PR without approval.

PR #92, opened 2022-05-19
add api for avs-agent to embed wall-clock timestamps

lzuosym:


mpeg2/src/pes.rs
```diff
@@ -38,6 +41,7 @@ impl<'a> Iterator for Packetize<'a> {
         let adaptation_field = self.header.map(|header| {
             let mut af = ts::AdaptationField {
                 random_access_indicator: if self.config.random_access_indicator { Some(true) } else { None },
+                temi_timeline_descriptors: Cow::Borrowed(&self.config.temi),
```
lzuosym, line 44: @ccbrown `temi` info is shared by all `ts:Packet`s produced by `pes::Packet`, there is not need to duplicate `temi` for `ts::Packet`, but I couldn't find a way to pass Rust lifetime check.

lzuosym: Thanks! I made simple things complicated!

mpeg2/src/pes.rs
```diff
@@ -13,6 +15,7 @@ pub struct PacketizationConfig {
     pub packet_id: u16,
     pub random_access_indicator: bool,
     pub continuity_counter: u8,
+    pub temi: Option<Vec<TEMITimelineDescriptor>>,
```
ccbrown, line 18: Is there any difference in behavior between `None` and `Some(vec![])`? Why not just make this a `Vec`?

```suggestion
    pub temi: Vec<TEMITimelineDescriptor>,
```

mpeg2/src/ts.rs
```diff
@@ -21,7 +21,7 @@ pub struct AdaptationField {
     pub random_access_indicator: Option<bool>,
     pub program_clock_reference_27mhz: Option<u64>,
     pub private_data_bytes: Vec<u8>,
-    pub temi_timeline_descriptors: Vec<TEMITimelineDescriptor>,
+    pub temi_timeline_descriptors: Option<Vec<TEMITimelineDescriptor>>,
```
ccbrown, line 24: Same question.

mpeg2/src/pes.rs
ccbrown, line 18: Also, for consistency:

```suggestion
    pub temi_timeline_descriptors: Vec<TEMITimelineDescriptor>,
```

lzuosym: By using `Option`, we can simply call `take()` to move the temi vectors into AdaptationField, all the following would be `None`.  
```rust
fn next(&mut self) -> Option<Self::Item> {
    let adaptation_field = self.header.map(|header| {
        let mut af = ts::AdaptationField {
            random_access_indicator: if self.config.random_access_indicator { Some(true) } else { None },
            temi_timeline_descriptors: self.config.temi.take(),
```

mpeg2/src/ts.rs
lzuosym: it can be removed.

mpeg2/src/pes.rs
lzuosym: just lazy, but should make it consistent

ccbrown: Let's try to avoid a leaky abstraction here. The implementation details of that function shouldn't dictate what the public API looks like.

There are a few ways you can do the same thing without making this an `Option`. For example, you could `mem::swap` it with an empty `Vec`.

ccbrown (changes requested):

mpeg2/src/pes.rs
lzuosym: Oh, okay, let me check the `mem::swap` method.

mpeg2/src/muxer.rs
```diff
@@ -196,6 +190,7 @@ pub struct Packet<'a> {
     pub random_access_indicator: bool,
     pub pts_90khz: Option<u64>,
     pub dts_90khz: Option<u64>,
+    pub temi: Vec<TEMITimelineDescriptor>,
```
ccbrown, line 193: I'm always a fan of spelling things out fully in `pub` APIs:

```suggestion
    pub temi_timeline_descriptor: Vec<TEMITimelineDescriptor>,
```

ccbrown (changes requested):
Naming suggestion. Otherwise, this looks good to me!

mpeg2/src/muxer.rs
lzuosym: will keep that in mind!!!!

ccbrown (approved):

PR #93, opened 2022-05-21
fix rust compilation error

lzuosym:
https://github.com/sportsball-ai/av-rs/runs/6533073661?check_suite_focus=true

### DIFF
--- a/mpeg2/src/interleaving_muxer.rs
+++ b/mpeg2/src/interleaving_muxer.rs
@@ -263,6 +263,7 @@ mod test {
             random_access_indicator: true,
             pts_90khz: Some(pts_90khz),
             dts_90khz: None,
+            temi_timeline_descriptors: vec![]
         }
     }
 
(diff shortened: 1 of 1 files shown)


mpeg2/src/interleaving_muxer.rs
```diff
@@ -263,6 +263,7 @@ mod test {
             random_access_indicator: true,
             pts_90khz: Some(pts_90khz),
             dts_90khz: None,
+            temi_timeline_descriptors: vec![]
```
ccbrown, line 266: Shouldn't there be a comma at the end of this line?

If you don't already, I'd highly recommend making sure you have an integration that runs rustfmt for you when you edit files.

lzuosym: adding comma after the last item seems to be  idiomatic rust, but my local clippy didn't complain

ccbrown: Yeah, Clippy is happy with it, but rustfmt isn't, and unfortunately we don't run rustfmt checks in CI yet. :-/

ccbrown (approved):

PR #95, opened 2022-06-02
[AVS-251] Add API for getting TEMI timeline descriptor data from mpegts-segmenter crate

lzuosym:
Add a TEMI timeline descriptor field to `SegmentInfo` struct to provide wallclock timestamps in segments


mpegts-segmenter/src/segmenter.rs
```diff
@@ -567,4 +574,31 @@ mod test {
             ]
         );
     }
+
+    #[tokio::test]
+    async fn test_segmenter_with_temi_timeline_descriptor() {
+        let mut storage = MemorySegmentStorage::new();
+        {
+            let mut f = File::open("src/testdata/temi_timeline-ntp-ts.ts").unwrap();
```
ccbrown, line 582:
```suggestion
            let mut f = File::open("src/testdata/temi-timeline-ntp-ts.ts").unwrap();
```

ccbrown (changes requested):

mpegts-segmenter/src/segmenter.rs
```diff
@@ -163,6 +167,7 @@ impl<S: SegmentStorage> Segmenter<S> {
                     pcr: self.pcr.unwrap_or(0),
                     pts: None,
                     bytes_written: 0,
+                    temi_timeline_descriptor,
```
ccbrown, line 170: I think you could do something like this to avoid unnecessary clones:

```suggestion
                    temi_timeline_descriptor: af.and_then(|af| af.temi_timeline_descriptors.first()),
```

ccbrown, line 170: This logic is slightly different from the PCR and PTS logic. Do you think it might be better to adjust it to match one of those? In other words, maybe it'd be good to set the timeline descriptor to `None` here, and set it below, around like 183 with the same sort of logic as the PTS.

lzuosym: `first()` returns and a reference, we need ownership.

lzuosym: I agree with you! 👍

PR #96, opened 2022-06-10
[AVS-15] add support to parse fragmented mp4 file

lzuosym:
Previously fragmented mp4 files are not able to parsed correctly, which causes the `av-service` to be unable to extract frames from fragmented mp4 files. This will most likely enable us to generate thumbnails from fragmented mp4 files.

### DIFF
--- a/qtff/src/file.rs
+++ b/qtff/src/file.rs
@@ -71,6 +71,8 @@ impl File {
     fn get_moof_data(&mut self) -> Result<Vec<MovieFragment>> {
         let mut fragments = vec![];
         loop {
+            let start_position = self.f.stream_position()?;
+
             let atom = match AtomReader::new(&mut self.f).find(|a| match a {
                 Ok(a) => a.typ == MovieFragment::TYPE,
                 Err(_) => true,
@@ -81,7 +83,10 @@ impl File {
             };
             let mut buf = Vec::new();
             atom.data(&mut self.f).read_to_end(&mut buf)?;
-            fragments.push(MovieFragment::read(Cursor::new(buf.as_slice()))?);
+
+            let mut movie_fragment = MovieFragment::read(Cursor::new(buf.as_slice()))?;
+            movie_fragment.fragment_header.start_position = start_position;
+            fragments.push(movie_fragment);
         }
     }
 
@@ -743,6 +748,7 @@ mod tests {
                 version: 0,
                 flags: 0,
                 sequence_number: 1,
+                start_position: 834,
             }
         );
         assert_eq!(fragment.track_fragments.len(), 1);
--- a/qtff/src/file.rs
+++ b/qtff/src/file.rs
@@ -14,7 +14,7 @@ use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 pub struct File {
     f: std::fs::File,
     movie_data: Option<MovieData>,
-    movie_fragments: Option<Vec<MovieFragment>>,
+    movie_fragments: Option<Vec<(u64, MovieFragment)>>,
 }
 
 enum Data {
@@ -68,7 +68,7 @@ impl File {
         }
     }
 
-    fn get_moof_data(&mut self) -> Result<Vec<MovieFragment>> {
+    fn get_moof_data(&mut self) -> Result<Vec<(u64, MovieFragment)>> {
         let mut fragments = vec![];
         loop {
             let start_position = self.f.stream_position()?;
@@ -84,13 +84,11 @@ impl File {
             let mut buf = Vec::new();
             atom.data(&mut self.f).read_to_end(&mut buf)?;
 
-            let mut movie_fragment = MovieFragment::read(Cursor::new(buf.as_slice()))?;
-            movie_fragment.fragment_header.start_position = start_position;
-            fragments.push(movie_fragment);
+            fragments.push((start_position, MovieFragment::read(Cursor::new(buf.as_slice()))?));
         }
     }
 
-    pub fn get_movie_fragment_data(&mut self) -> Result<Vec<MovieFragment>> {
+    pub fn get_movie_fragments(&mut self) -> Result<Vec<(u64, MovieFragment)>> {
         if self.movie_data.is_none() {
             self.get_movie_data()?;
         }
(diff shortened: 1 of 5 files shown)


lzuosym:
@ccbrown I don't know why `xcoder` build  is failing. I cannot reproduce it since it needs to be run in linux environment. Please take a look.

qtff/src/file.rs
```diff
@@ -65,6 +68,42 @@ impl File {
...
+            let mut buf = Vec::new();
+            atom.data(&mut self.f).read_to_end(&mut buf)?;
+
+            let mut movie_fragment = MovieFragment::read(Cursor::new(buf.as_slice()))?;
+            movie_fragment.fragment_header.start_position = start_position;
+            fragments.push(movie_fragment);
+        }
+    }
+
+    pub fn get_movie_fragment_data(&mut self) -> Result<Vec<MovieFragment>> {
```
ccbrown, line 93: Naming suggestion since this returns a collection of `MovieFragment`s:

```suggestion
    pub fn get_movie_fragments(&mut self) -> Result<Vec<MovieFragment>> {
```

qtff/src/moof.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+impl AtomData for MovieFragment {
+    const TYPE: FourCC = FourCC::MOOF;
+}
+
+impl ReadData for MovieFragment {
+    fn read<R: Read + Seek>(mut reader: R) -> Result<Self> {
+        Ok(Self {
+            fragment_header: read_one(&mut reader)?.ok_or(Error::MalformedFile("missing movie header"))?,
```
ccbrown, line 188:
```suggestion
            fragment_header: read_one(&mut reader)?.ok_or(Error::MalformedFile("missing movie fragment header"))?,
```

qtff/src/moof.rs
```diff
@@ -0,0 +1,192 @@
...
+            version,
+            flags: reader.read_u24::<BigEndian>()?,
+            base_media_decode_time: if version == 1 {
+                reader.read_u8()? as u32
+            } else {
+                reader.read_u32::<BigEndian>()?
+            },
+        })
+    }
+}
```
ccbrown, line 88: It looks like this isn't used. Since it also looks untested, I'd say just remove it for now.

qtff/src/moof.rs
```diff
@@ -0,0 +1,192 @@
...
+use crate::{read_all, read_one, AtomData, FourCC, ReadData};
+use byteorder::{BigEndian, ReadBytesExt};
+use std::io::{Read, Seek};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct FragmentHeader {
+    pub version: u8,
+    pub flags: u32,
+    pub sequence_number: u32,
+    pub start_position: u64,
```
ccbrown, line 11: This field is kinda special. It's not actually used by `FragmentHeader` or read from the file.

Maybe it would make more sense to remove it from here and have `get_movie_fragment_data` return the position separately like by returning `Vec<(u64, MovieFragment)>`.

ccbrown (commented):
I've looked through the spec and as far as I can tell everything here is correct. I just have a few cleanliness comments.

qtff/src/moof.rs
lzuosym: You're right! it should be a part of `FragmentHeader`. Putting it here just because I don't know where it should go. It makes more sense to return a tuple. 👍

qtff/src/moof.rs
lzuosym: good catch! We don't need it in this PR, will remove it.

ccbrown:
@epingel2 any idea on the xcoder failure?

epingel2:
It looks like It is using bindgen 0.61.0 which still had this bug https://github.com/rust-lang/rust-bindgen/issues/2218 which causes unit tests to overflow their stack. Previously we had been using an older version of Bingen which doesnt have this issue.

xcoder/xcoder-sys/Cargo.toml
```diff
@@ -6,5 +6,5 @@ build = "build.rs"
 
 [build-dependencies]
 # We're very permissive here with bindgen due to https://github.com/rust-lang/cargo/issues/5237
-bindgen = "0.*"
+bindgen = "0.59.2"
```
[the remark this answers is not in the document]
lzuosym: I saw the comment above, just want to confirm it's the new bindgen version that breaks test. I don't know if there is any alternative way to solve this. It's an issue introduced in bindgen 0.60.0, they claimed it be fixed in 0.60.1, but seems to be not.
https://github.com/rust-lang/rust-bindgen/issues/2218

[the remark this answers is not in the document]
ccbrown: I think I'd prefer to just disable the layout tests for this crate:

```diff
diff --git a/xcoder/xcoder-sys/Cargo.toml b/xcoder/xcoder-sys/Cargo.toml
index 60cf1e5..f72b38d 100644
--- a/xcoder/xcoder-sys/Cargo.toml
+++ b/xcoder/xcoder-sys/Cargo.toml
@@ -6,5 +6,5 @@ build = "build.rs"
 
 [build-dependencies]
 # We're very permissive here with bindgen due to https://github.com/rust-lang/cargo/issues/5237
-bindgen = "0.59.2"
+bindgen = "0.*"
 cc = "1.0"
diff --git a/xcoder/xcoder-sys/build.rs b/xcoder/xcoder-sys/build.rs
index a2ddc6a..4ce2e93 100644
--- a/xcoder/xcoder-sys/build.rs
+++ b/xcoder/xcoder-sys/build.rs
@@ -68,6 +68,9 @@ fn main() {
             .whitelist_function("ni_.+")
             .whitelist_type("ni_.+")
             .whitelist_var("NI_.+")
+            // bindgen 0.60.0 broke layout tests: https://github.com/rust-lang/rust-bindgen/issues/2218
+            // 0.60.1 claims to have fixed the issue, but does not.
+            .layout_tests(false)
             .clang_arg(format!("-I{}", source_path.to_str().unwrap()))
             .generate()
             .expect("unable to generate bindings");
```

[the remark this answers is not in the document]
lzuosym: 👍 we can re-enable the `layout_tests` after bindgen fix the stack overflow.

ccbrown (approved):

PR #97, opened 2022-06-24
[AVS-286] Add resource reservation for transcode pool

epingel2:
This PR adds the ability to calculate the load and reserve resources for an entire transcode job at once and ensure that it will be allocated to one device. This will be prerequisite for a change in the av-service which will implement these changes.

### DIFF
--- a/xilinx/src/xlnx_encoder.rs
+++ b/xilinx/src/xlnx_encoder.rs
@@ -9,7 +9,6 @@ pub struct XlnxEncoder {
 
 impl XlnxEncoder {
     pub fn new(xma_enc_props: &mut XmaEncoderProperties, xlnx_enc_ctx: &mut XlnxEncoderXrmCtx) -> Result<Self, SimpleError> {
-        
         let enc_session = xlnx_create_enc_session(xma_enc_props, xlnx_enc_ctx)?;
 
         let out_buffer = unsafe { xma_data_buffer_alloc(0, true) };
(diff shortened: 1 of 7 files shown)


xilinx/src/xlnx_encoder.rs
```diff
@@ -161,6 +159,7 @@ mod encoder_tests {
         let xrm_ctx = unsafe { xrmCreateContext(XRM_API_VERSION_1) };
 
         let cu_list_res: xrmCuListResource = Default::default();
+        let mut _enc_count = 0;
```
ccbrown, line 162: Is this intentional?

epingel2: No this is accidental.

xilinx/src/xlnx_transcoder_utils.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub scal_load: i32,
+    pub enc_load: i32,
+    pub enc_num: i32,
+}
+
+pub struct XlnxTranscodeXrmCtx {
+    pub xrm_ctx: xrmContext,
+    pub transcode_load: XlnxTranscodeLoad,
+    pub reserve_idx: u64,
+}
```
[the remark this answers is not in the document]
epingel2: That's a fair point. At the moment, the caller creates those fields, but It would make more sense to have a new function perform those functions.

ccbrown (approved):

PR #102, opened 2022-07-19
[AVS-288] Update the video encoder abstraction 

lzuosym:
This adds a new parameter `EncodedFrameType` to `VideoEncoder::encode` API to allows users to explicitly specify encoded frame types for libx264 or Video Toolbox.

1. For x264 encoder, there is a direct mapping from `EncodedFrameType` to properties defined in x264
2. For Video Toolbox, [`kVTEncodeFrameOptionKey_ForceKeyFrame`](https://developer.apple.com/documentation/videotoolbox/kvtencodeframeoptionkey_forcekeyframe) property seems to be used to force the current encoded frame to be a key frame. But there are not corresponding properties for `Auto` and `Predicted`.  [`kVTEncodeFrameOptionKey_BaseFrameQP`](https://developer.apple.com/documentation/videotoolbox/kvtencodeframeoptionkey_baseframeqp) is another frame encoding option, but it's not clear what it is and how to use it.

### DIFF
--- a/macos/video-toolbox/src/compression_session.rs
+++ b/macos/video-toolbox/src/compression_session.rs
@@ -112,7 +112,7 @@ impl<C: Send> CompressionSession<C> {
                     EncodedFrameType::Key => {
                         frame_options.set_value(sys::kVTEncodeFrameOptionKey_ForceKeyFrame as _, Boolean::from(true).cf_type_ref() as _);
                     }
-                    EncodedFrameType::Auto | EncodedFrameType::Predicted => {}
+                    EncodedFrameType::Auto => {}
                 };
 
                 sys::VTCompressionSessionEncodeFrame(
--- a/macos/video-toolbox/src/compression_session.rs
+++ b/macos/video-toolbox/src/compression_session.rs
@@ -56,7 +56,7 @@ impl<C: Send> CompressionSession<C> {
                 Err(status.into())
             } else {
                 result(status.into()).map(|_| CompressionSessionOutputFrame {
-                    sample_buffer: unsafe { SampleBuffer::with_cf_type_ref(sample_buffer as _) },
+                    sample_buffer: SampleBuffer::with_cf_type_ref(sample_buffer as _),
                     context: frame_context,
                 })
             });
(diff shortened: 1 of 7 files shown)


macos/video-toolbox/src/compression_session.rs
```diff
@@ -102,10 +103,20 @@ impl<C: Send> CompressionSession<C> {
...
+    pub fn encode_frame(&mut self, image_buffer: ImageBuffer, presentation_time: Time, context: C, frame_type: EncodedFrameType) -> Result<(), OSStatus> {
         let context = Box::new(context);
         result(
             unsafe {
+                let frame_properties = match frame_type {
+                    EncodedFrameType::Key => {
+                        let mut md = MutableDictionary::new_cf_type();
+                        md.set_value(sys::kVTEncodeFrameOptionKey_ForceKeyFrame as _, Boolean::from(true).cf_type_ref() as _);
+                        let d = Dictionary::from(md);
+                        d.cf_type_ref()
```
lzuosym, line 115: @ccbrown there seems to be something wrong here, which fails the unit test for VideoToolbox encoder, but I couldn't figure it out, can you please take a look?

macos/video-toolbox/src/compression_session.rs
```diff
@@ -102,10 +103,18 @@ impl<C: Send> CompressionSession<C> {
...
         let context = Box::new(context);
         result(
             unsafe {
+                let mut frame_options = MutableDictionary::new_cf_type();
+                match frame_type {
+                    EncodedFrameType::Key => {
+                        frame_options.set_value(sys::kVTEncodeFrameOptionKey_ForceKeyFrame as _, Boolean::from(true).cf_type_ref() as _);
+                    }
+                    EncodedFrameType::Auto | EncodedFrameType::Predicted => {}
+                };
```
ccbrown, line 116: The dictionary was being freed before the `VTCompressionSessionEncodeFrame` call, resulting in a crash. I refactored things a bit to make this work.

ccbrown (changes requested):
I went ahead and update the xcoder crates too. Let's remove `Predicted` for now though. Otherwise this lgtm.

macos/video-toolbox/src/compression_session.rs
lzuosym: That makes sense! 👍

ccbrown (approved):

PR #104, opened 2022-07-26
add rsrc_init api

ccbrown:
Adds an API to execute the "init_rsrc" step that has to be done on machines using NETINT hardware.


raven-txm (approved):
lgtm

PR #105, opened 2022-07-30
fix xcoder-logan memory leak

ccbrown:
See the test and the comment. Before this PR, the encoder would abort every 6000 frames.


beiriannydd:
Good spot.  Thanks for the test.

lzuosym (approved):

PR #109, opened 2022-08-04
add gitattributes

ccbrown:
Not sure why git lfs behaves differently on Linux, but this seems to be necessary to keep it from thinking the .yuv file is modified.


lzuosym (approved):
Most likely your Linux environment doesn't have a `.gitconfig` file in your home directory, or the `[filter "lfs"]` section is missing in the .gitconfig file

ccbrown:
Yep, you're right.

PR #110, opened 2022-08-05
refactor xilinx props

ccbrown:
This refactors the way creating properties for the decoder, scaler, and encoder works so that the user doesn't have to box things or worry about delicate lifetime requirements.

I don't think there was any bug here or in our code which used this, but this makes things safer and simpler to users of the crate.


epingel2 (approved):
This looks very good. It is a large improvement! Makes it a lot easier to understand and to work with. I don't see any issues.

PR #112, opened 2022-08-11
[AVS-373] use pkg-config to find kvazaar lib

scottlamb:


kvazaar-sys/build.rs
```diff
@@ -1,17 +1,15 @@
-extern crate bindgen;
-
 use std::env;
 use std::path::PathBuf;
 
 fn main() {
-    println!("cargo:rustc-link-lib=kvazaar");
-
+    let library = pkg_config::probe_library("kvazaar").unwrap();
```
ccbrown, line 5: I see you're using this for include paths, but what about the linker search path? And shouldn't `cargo:rustc-link-lib=kvazaar` still be given so rustc knows it needs to link to kvazaar?

```suggestion
    let library = pkg_config::probe_library("kvazaar").unwrap();
    for path in library.link_paths {
        println!("cargo:rustc-link-search={}", path.display());
    }
    println!("cargo:rustc-link-lib=kvazaar");
```

scottlamb: `pkg_config::probe_library` does that for us:

https://github.com/rust-lang/pkg-config-rs/blob/4246c5a3ef4acba7eb3d37a47669e187df100711/src/lib.rs#L610

ccbrown: Well, that's handy. :)

ccbrown (approved):

PR #113, opened 2022-08-11
run xcoder tests in ci

ccbrown:
We now have a few machines in the data center loaded up with Quadras. This PR uses them to run xcoder-quadra tests. And it uses the pre-existing machines to run xcoder-logan tests.


raven-txm (approved):
I like nice tests and I cannot lie

PR #115, opened 2022-08-17
[AVS-379] quadra cropping and scaling

ccbrown:
This updates the xcoder-quadra crate to use SW release 4.0 and adds hardware cropping and scaling functionality behind a nice, simple, safe API.

I'd recommend starting by looking at the usage in `test_transcoding` and `test_8k_ladder`.


xcoder/xcoder-quadra/src/scaler.rs
```diff
@@ -0,0 +1,217 @@
...
+            });
+
+            (**session).session_id = sys::NI_INVALID_SESSION_ID;
+            (**session).device_handle = sys::NI_INVALID_DEVICE_HANDLE;
+            (**session).blk_io_handle = sys::NI_INVALID_DEVICE_HANDLE;
+            (**session).hw_id = config.hardware.id;
+            (**session).sender_handle = config.hardware.device_handle;
+            (**session).device_type = sys::ni_device_type_t_NI_DEVICE_TYPE_SCALER as _;
+            (**session).scaler_operation = sys::_ni_scaler_opcode_NI_SCALER_OPCODE_SCALE;
+            (**session).keep_alive_timeout = sys::NI_DEFAULT_KEEP_ALIVE_TIMEOUT;
```
lzuosym, line 48: same, can we use `ni_device_session_context_init ` to initialize the session context?

lzuosym (commented):
LGTM.  Just a few questions.

In Github CI why do we get `ni_dec_fme_buffer_pool_free: NOT allocated` in most of the quadra unit tests? 

https://github.com/sportsball-ai/av-rs/runs/7875049570?check_suite_focus=true

xcoder/xcoder-quadra/src/decoder.rs
```diff
@@ -428,8 +463,8 @@ mod test {
         let mut encoded = vec![];
 
         while !decoder.is_finished() {
-            if let Some(mut frame) = decoder.try_read_decoded_frame().unwrap() {
-                if let Some(mut output) = unsafe { encoder.encode_data_io((), frame.as_data_io_mut_ptr()).unwrap() } {
+            if let Some(frame) = decoder.try_read_decoded_frame().unwrap() {
+                if let Some(mut output) = encoder.encode_hardware_frame((), frame.into()).unwrap() {
```
scottlamb, line 467: Similar question: when should a hardware frame not produce an output frame? or is this just when the accelerator isn't done producing it? if the latter, should we be looping here to catch up on subsequent frames?

scottlamb (commented):
Couple basic questions to start.

xcoder/xcoder-quadra/src/cropper.rs
```diff
@@ -0,0 +1,226 @@
...
+
+            Ok(Self {
+                session: ScopeGuard::into_inner(session),
+                did_initialize: false,
+            })
+        }
+    }
+
+    pub fn crop(&mut self, f: &XcoderHardwareFrame, crop: XcoderCrop) -> Result<XcoderHardwareFrame> {
+        const PIXEL_FORMAT: i32 = 0x103; // GC620_I420;
```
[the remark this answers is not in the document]
ccbrown: It's not quite as straightforward as that since GC620_I420 is preprocessor define instead of a constant. See https://github.com/rust-lang/rust-bindgen/issues/316.

But I did go ahead and add something.

xcoder/xcoder-quadra/src/decoder.rs
```diff
@@ -428,8 +463,8 @@ mod test {
         let mut encoded = vec![];
 
         while !decoder.is_finished() {
-            if let Some(mut frame) = decoder.try_read_decoded_frame().unwrap() {
-                if let Some(mut output) = unsafe { encoder.encode_data_io((), frame.as_data_io_mut_ptr()).unwrap() } {
+            if let Some(frame) = decoder.try_read_decoded_frame().unwrap() {
```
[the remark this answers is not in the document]
ccbrown: This originated from my code for the previous generation of NETINT hardware [here](https://github.com/sportsball-ai/av-service/blob/cdf3370b0b23052f6c6e19d35915946462d547cd/microservices/netint-transcoding-service/src/decoder.rs) (which is being obsoleted). The second guess is closer, except as far as I can remember, there was no limitation, just how it ended up. Probably just a less-than-idiomatic translation of NETINT's sample code.

Refactored so the caller can just use a `while let` loop and doesn't have to use `is_finished`.

xcoder/xcoder-quadra/src/decoder.rs
ccbrown: Encoders in general often delay the output by a few frames so they can do things like re-order the output with B-frames when it makes sense. We can't expect to always be able to pull out as many frames as we've sent to the encoder. This is why after we're done decoding, we send a flush signal to the encoder and loop over the remaining frames. So I don't see any obvious changes that should be made here.

xcoder/xcoder-quadra/src/decoder.rs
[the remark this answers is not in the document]
scottlamb: Ahh, that's nicer, thanks. Can we drop the `pub` from `fn try_read_decoded_frame` now?

xcoder/xcoder-quadra/src/decoder.rs
```diff
@@ -257,6 +275,16 @@ impl<E: Error, I: XcoderDecoderInput<E>> XcoderDecoder<I, E> {
...
+        while !self.is_finished() {
+            if let Some(frame) = self.try_read_decoded_frame()? {
+                return Ok(Some(frame));
+            }
+        }
+        Ok(None)
+    }
+
     /// Tries to read a decoded frame. If none is returned and `is_finished` returns false, the caller should try again later.
     pub fn try_read_decoded_frame(&mut self) -> Result<Option<XcoderDecodedFrame>, XcoderDecoderError<E>> {
```
ccbrown, line 289:
```suggestion
    fn try_read_decoded_frame(&mut self) -> Result<Option<XcoderDecodedFrame>, XcoderDecoderError<E>> {
```

xcoder/xcoder-quadra/src/decoder.rs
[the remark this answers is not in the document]
ccbrown: Sure.

xcoder/xcoder-quadra/src/decoder.rs
scottlamb: Ahh, I see this in the libxcoder docs, section 6.6.1, matching what you just said. "The NETINT encoder engine may buffer several frames before producing encoded packets depending on the stream’s GOP structure which may contain out of sequence frames." And they have an example similar to this which reads at most one frame after each write of one frame. Maybe a comment would help future readers (such as future-me after I forget this ;-)).

So basically, the engine picks some N. First N frames of input produce no output then each following frame produces 1 (from N frames ago), then at flush it produces the last N. There's no possibility that we will fall more than N frames with this loop.

lzuosym (approved):

scottlamb (approved):

ccbrown:
> LGTM. Just a few questions.
> 
> In Github CI why do we get `ni_dec_fme_buffer_pool_free: NOT allocated` in most of the quadra unit tests?
> 
> https://github.com/sportsball-ai/av-rs/runs/7875049570?check_suite_focus=true

Sorry, forgot to answer this one. `ni_dec_fme_buffer_pool_free` gets called unconditionally in `ni_decoder_session_close`. But we don't create a frame buffer since we're using hardware-only decoding. I'm not sure why they decided to log that message as an error, but it doesn't appear to be a problem or avoidable unless we want to create a frame buffer just to make that go away.

lzuosym:
> Sorry, forgot to answer this one. `ni_dec_fme_buffer_pool_free` gets called unconditionally in `ni_decoder_session_close`. But we don't create a frame buffer pool since we're using hardware-only decoding. I'm not sure why they decided to log that message as an error, but it doesn't appear to be a problem or avoidable unless we want to create a frame buffer pool just to make that go away.

Agree! This appears not to be a problem. Thank you for looking into it.

PR #122, opened 2022-08-31
add h264 vui and sei encoding

ccbrown:
This adds encoding of VUI parameters and SEI pic timing messages to the h264 crate.

To help make things nicer, it also derives more and makes `RBSP` an `IntoIterator`.


lzuosym (approved):
I looked through the syntax for vui and sei in ITU-T H.264 (04/2017), everything here looks good to me.

PR #123, opened 2022-09-03
encode timecodes with xcoder-logan

ccbrown:
This allows video frames to optionally provide a timecode, which will be written when encoding H264 with xcoder-logan.


lzuosym (approved):
LGTM

PR #127, opened 2022-09-10
Avs 488 video toolboox keyframe settings

lzuosym:
By default, Video Toolbox inserts a keyframe every 1 second. In HA mode, we insert keyframes every 3 seconds, this will keep Video Toolbox from inserting additional keyframes in HA mode.

### DIFF
--- a/macos/video-toolbox/src/video_encoder.rs
+++ b/macos/video-toolbox/src/video_encoder.rs
@@ -27,6 +27,7 @@ pub struct VideoEncoderConfig {
     pub fps: f64,
     pub codec: VideoEncoderCodec,
     pub input_format: VideoEncoderInputFormat,
+    pub high_availability_mode: bool,
 }
 
 #[derive(Clone, Debug)]
@@ -63,6 +64,9 @@ impl<F: Send> VideoEncoder<F> {
             sess.set_property(sys::kVTCompressionPropertyKey_AllowFrameReordering, Boolean::from(false))?;
             sess.set_property(sys::kVTCompressionPropertyKey_RealTime, Boolean::from(true))?;
             sess.set_property(sys::kVTCompressionPropertyKey_ExpectedFrameRate, Number::from(config.fps))?;
+            if config.high_availability_mode {
+                sess.set_property(sys::kVTCompressionPropertyKey_MaxKeyFrameInterval, Number::from(i32::MAX))?;
+            }
 
             match &config.codec {
                 VideoEncoderCodec::H264 { bitrate } => {
@@ -254,6 +258,7 @@ mod test {
             fps: 29.97,
             codec,
             input_format: VideoEncoderInputFormat::Yuv420Planar,
+            high_availability_mode: false,
         })
         .unwrap();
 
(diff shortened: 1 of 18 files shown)


macos/video-toolbox/src/video_encoder.rs
```diff
@@ -27,6 +27,7 @@ pub struct VideoEncoderConfig {
     pub fps: f64,
     pub codec: VideoEncoderCodec,
     pub input_format: VideoEncoderInputFormat,
+    pub high_availability_mode: bool,
```
ccbrown, line 30: In this repo, "high availability" is meaningless. I'd suggest just making this an optional max key frame interval:

```suggestion
    pub max_keyframe_interval: Option<i32>,
```

lzuosym: 👍

macos/video-toolbox/src/video_encoder.rs
```diff
@@ -300,15 +305,57 @@ mod test {
...
+            };
+            if i % 90 == 0 {
+                if let Some(output) = encoder.encode(frame, EncodedFrameType::Key).unwrap() {
+                    keyframe_count += if output.encoded_frame.is_keyframe { 1 } else { 0 };
+                }
+            } else if let Some(output) = encoder.encode(frame, EncodedFrameType::Auto).unwrap() {
+                keyframe_count += if output.encoded_frame.is_keyframe { 1 } else { 0 };
+            }
+        }
+        assert_eq!(keyframe_count, 12);
```
ccbrown, line 340: This seems likely to break in future macOS updates if the default encoder settings change. I'm not sure we really care how many key frames there are in this case, so it might make sense to loosen up or remove this check.

Maybe just make the first frame `Key` and the rest `Auto` and the assert that `keyframe_count` > 1?

macos/video-toolbox/src/video_encoder.rs
```diff
@@ -300,15 +305,57 @@ mod test {
...
+            let frame = TestFrame {
+                samples: vec![y, u.clone(), v.clone()],
+            };
+            if i % 90 == 0 {
+                if let Some(output) = encoder.encode(frame, EncodedFrameType::Key).unwrap() {
+                    keyframe_count += if output.encoded_frame.is_keyframe { 1 } else { 0 };
+                }
+            } else if let Some(output) = encoder.encode(frame, EncodedFrameType::Auto).unwrap() {
+                keyframe_count += if output.encoded_frame.is_keyframe { 1 } else { 0 };
+            }
```
ccbrown, line 338: Could be a little more DRY:

```suggestion
            if let Some(output) = encoder.encode(frame, if i % 90 == 0 { EncodedFrameType::Key } else { EncodedFrameType::Key }).unwrap() {
                keyframe_count += if output.encoded_frame.is_keyframe { 1 } else { 0 };
            }
```

macos/video-toolbox/src/video_encoder.rs
lzuosym: Good point!

macos/video-toolbox/src/video_encoder.rs
lzuosym: Ah, very neat! Thanks.

ccbrown (approved):

PR #128, opened 2022-09-12
use pkg-config include paths for x264

scottlamb:
This allows building with no environment variable overrides on macOS with Homebrew-installed x264.

Same idea as <https://github.com/sportsball-ai/av-rs/pull/112>, except the include paths are passed to both cc and clang/bindgen.

Somehow the build was already working on macOS on CI (I don't get it?!?), but it didn't work on my laptop. It does now.


epingel2 (approved):
Nice. This is a good change. lgtm

lzuosym (approved):

PR #129, opened 2022-09-16
add ffmpeg crate

ccbrown:
This moves the FFmpeg AVIO structures I recently iterated on into a new crate where they can be re-used, as I think the rule of 3 applies here. Really they should probably be used anywhere we use FFmpeg from Rust.


ffmpeg/src/avio.rs
```diff
@@ -0,0 +1,241 @@
...
+        Writer::new(&mut *buf).unwrap();
+
+        let mut cur = std::io::Cursor::new(vec![]);
+        Writer::new_seekable(&mut cur).unwrap();
+    }
+
+    #[test]
+    fn test_lifetimes() {
+        let t = trybuild::TestCases::new();
+        t.compile_fail("tests/avio/bad_reader_lifetime.rs");
```
scottlamb, line 238: Neat! I've never used this before.

ffmpeg/README.md
```diff
@@ -0,0 +1,3 @@
+# ffmpeg
+
+This crate contains highish level facilities for working with FFmpeg libraries.
```
scottlamb, line 3: Seems worth some mention here (or in an ADR) of why write a new one vs use/contribute to/fork existing libraries. I think the most popular these days is [`ffmpeg-next`](https://crates.io/crates/ffmpeg-next), a fork of the `ffmpeg` crate (and same forker as `ffmpeg-sys-next`, see other comment).

FYI, I also wrote my Rust ffmpeg wrapper crate, [`moonfire-ffmpeg`](https://crates.io/crates/moonfire-ffmpeg). I don't think most of my reasons (listed in the `README.md`) exactly apply here/today vs `ffmpeg-next` or that it'd necessarily make sense for txm to use my lib, but feel free to steal anything. My logging, dictionary, and error code [here](https://github.com/scottlamb/moonfire-ffmpeg/blob/master/src/avutil.rs) might be handy.

ffmpeg/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "ffmpeg"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+ffmpeg-sys = {version = "4", git = "https://github.com/meh/rust-ffmpeg-sys", default-features = false, features = ["avcodec", "avformat"]}
```
scottlamb, line 7: Did you consider moving to a more modern fork, [`ffmpeg-sys-next`](https://crates.io/crates/ffmpeg-sys-next), that compiles with ffmpeg 5? I think we should sometime, maybe not now. TODO?

ffmpeg/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#[macro_use]
+extern crate thiserror;
+
+pub mod avio;
```
scottlamb, line 4: Might be nice to organize by ffmpeg library, e.g. making this `ffmpeg::avformat::avio` rather than `ffmpeg::avio`, for a couple reasons:
* If this library grows and supports `avformat` vs `avcodec` vs `swscale` and all, ffmpeg libraries : Cargo features : module directories could be 1:1:1 so there's an easy way for our apps using this crate to pull in just the ffmpeg libraries they need.
* also makes it easier to flip between this and ffmpeg docs/include files.

ffmpeg/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#[macro_use]
+extern crate thiserror;
+
```
scottlamb, line 3: Maybe a `pub use ffmpeg_sys as sys` so that callers can just use same version this crate chose without having to futz with their own `Cargo.toml`/`Cargo.lock`? Seems nice to have until/unless this crate wraps all the ffmpeg APIs used by any of our software.

ccbrown: Yep, done.

ffmpeg/Cargo.toml
ccbrown: Yeah that'd be a nice TODO.

ffmpeg/src/lib.rs
ccbrown: Done.

ffmpeg/README.md
ccbrown: I tried a while back to use ffmpeg-next and it didn't have everything we needed. It's also in "maintenance mode", which I take to mean new features or accepted feature contributions should be expected.

ccbrown: I made a slight addition to the README here, but I don't know if this is the place to explain that.

Also, FYI there is precedence for us using staff libraries (e.g. almost all our Go codebases use [https://github.com/ccbrown/keyvaluestore](https://github.com/ccbrown/keyvaluestore)). While moonfire-ffmpeg doesn't look like it would quite cover all our needs at the moment, if it were developed further, TXM using it would certainly be a possibility.

scottlamb: Not surprised it doesn't have everything we need. But we could fork, and it also says though "If you have significant, demonstrable experience in Rust and multimedia-related programming, please let me know, I'll be more than happy to invite you as a collaborator. " I'd say we qualify. So I think the question is if it's a useful base to build on or not. (Part of that decision: I'm not sure about txm policy on licenses such as WTFPL.)

scottlamb: > Also, FYI there is precedence for us using staff libraries (e.g. almost all our Go codebases use [github.com/ccbrown/keyvaluestore](https://github.com/ccbrown/keyvaluestore)). While moonfire-ffmpeg doesn't look like it would quite cover all our needs at the moment, if it were developed further, TXM using it would certainly be a possibility.

Good to know, thanks. Off-hand though I'm not sure the goals of a library for Moonfire and the goals of a library need for TXM match well enough though: in particular `av-rs` is all-in on bindgen (where convenient comprehensive C bindings are helpful), and I don't think switching to that would make sense for Moonfire (where I don't care so much about that want compilation to be really really easy and fast).

ccbrown: One of the other problems is actually just that we do certain things that we can't even do with FFmpeg. Hence the existence of the low level parsing crates in this repo. That makes it really difficult to use any high level crate.

ccbrown: Also I just don't like FFmpeg, but maybe that's just me. I'd rather eliminate all our FFmpeg dependencies than invest in them.

scottlamb: > One of the other problems is actually just that we do certain things that we can't even do with FFmpeg. Hence the existence of the low level parsing crates in this repo. That makes it really difficult to use any high level crate.

As in: impedance mismatch between the abstractions those crates build and the ones in this crate for doing stuff directly?

> Also I just don't like FFmpeg, but maybe that's just me. I'd rather eliminate all our FFmpeg dependencies than invest in them (We really don't have very many fortunately.).

I'm not a huge ffmpeg fan either, and if it's realistic for TXM to not use it at all rather than invest in it much, I'm all for it. (Moonfire also abandoned ffmpeg for RTSP handling, but I don't have much hope of doing on-NVR analytics without it given that I want to support software decoding and don't really want to rewrite the 30,000+ lines of code in `libavcodec/*h26*.[ch]`.)

Those two reasons make sense to me; can you stick them in the README so it's clear to future folks? (future txm folks, anyone else who stumbles across `av-rs/ffmpeg` and might hope for something with more long-term commitment)

ccbrown: How does that look?

scottlamb (approved):

scottlamb (approved):

lzuosym (approved):

PR #130, opened 2022-09-17
ffmpeg error wrapper

scottlamb:
It's frustrating to see failures without the error message, so let's make propagating that easy.

E.g., this old code:

```rust
if av... < 0 {
    bail!("av... failed");
}
```

can become this:

```rust
if let Err(e) = ffmpeg::Error::wrap(av...) {
    bail!("av... failed: {}", e);
}
```

or `ffmpeg::Error::wrap(av...).map_err(...)?`.

This is mostly lifted from my similar code here:

impl https://github.com/scottlamb/moonfire-ffmpeg/blob/ab1fad1c9ed6c9d37f6c39135f797f3a0fa91941/src/avutil.rs#L262-L323 test https://github.com/scottlamb/moonfire-ffmpeg/blob/ab1fad1c9ed6c9d37f6c39135f797f3a0fa91941/src/avutil.rs#L425-L449

It's slightly simpler because callers can just use `ffmpeg_sys::AVERROR_*` instead of this code exposing the constants as I did in my `moonfire-ffmpeg` crate. I just made the inner int pub too.

### DIFF
--- a/ffmpeg/src/avutil/mod.rs
+++ b/ffmpeg/src/avutil/mod.rs
@@ -1,7 +1,6 @@
 use std::{ffi::CStr, os::raw::c_int};
 
 #[derive(Copy, Clone, PartialEq, Eq)]
-#[repr(transparent)]
 pub struct Error(pub c_int);
 
 impl Error {
--- a/ffmpeg/src/avutil/mod.rs
+++ b/ffmpeg/src/avutil/mod.rs
@@ -1,0 +1,62 @@
+use std::{ffi::CStr, os::raw::c_int};
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct Error(pub c_int);
+
+impl Error {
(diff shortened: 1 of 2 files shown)


ccbrown (approved):
👍

ffmpeg/src/avutil/mod.rs
```diff
@@ -0,0 +1,62 @@
+use std::{ffi::CStr, os::raw::c_int};
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+#[repr(transparent)]
```
lzuosym, line 4: We probably don't need this attribute `#[repr(transparent)]` on `Error`. Seems like this new type won't be used in any place that will cause ABI mismatch issue, or did I miss something?

scottlamb: You're right. It's not needed. Removed.

ccbrown (approved):

lzuosym (approved):

PR #133, opened 2022-09-27
support xcoder logan firmware 3.1.0

ccbrown:
This adds support for NETINT T4XX firmware 3.1.0.

In 3.1.0, all functions have names and types starting with "ni_logan_". This is so applications can link to both Logan and Quadra libraries simultaneously. However, this change can also be conveniently leveraged to allow applications to support 2.X and 3.X firmwares simultaneously. This PR gates this ability behind the "v2-compat" feature flag. The expectation is that this flag will be used to assist with the upgrades, but eventually users of the crate will prefer to support Logan v3 + Quadras instead of Logan v3 + v2.

To test this, I've added a new runner with the latest firmware. I've labeled it netint-t408-310X2013 and I've added a netint-t408-251X1E09 label to the old one. This repo now runs tests against both.

This will enable us to update NETINT firmware, and is very likely to resolve https://epf.atlassian.net/browse/PROD-572.

## How to Review

* Don't look at xcoder-logan-259-sys. It was renamed without changes from xcoder-logan-sys.
* Don't look at xcoder-logan-310-sys _except_ for build.rs and src/header.h. Otherwise this is the same as xcoder-logan-259-sys but with release 3.1.0 vendored.
* Review the changes in xcoder/xcoder-logan/src/encoder/mod.rs and xcoder/xcoder-logan/src/encoder/xcoder_259.rs. There was some refactoring to make a few bits re-usable across 2.5.9 and 3.1.0, and most importantly, this changes the way SEIs are inserted, so take an especially good look at `assemble_access_unit` and its usage.
* Don't look at xcoder/xcoder-logan/src/encoder/xcoder_310.rs. It is identical to xcoder/xcoder-logan/src/encoder/xcoder_259.rs except is uses "ni_logan_" functions from xcoder-logan-310-sys instead of "ni_" functions from xcoder-logan-259-sys.
* Review all the Cargo files.


Cargo.toml
```diff
@@ -26,7 +26,8 @@ members = [
     "kvazaar-sys",
     "xcoder",
     "xcoder/xcoder-logan",
-    "xcoder/xcoder-logan/xcoder-logan-sys",
+    "xcoder/xcoder-logan/xcoder-logan-259-sys",
+    "xcoder/xcoder-logan/xcoder-logan-310-sys",
```
raven-txm, line 30: 👍

xcoder/xcoder-logan/src/encoder/xcoder_259.rs
```diff
@@ -420,7 +309,7 @@ impl<F: RawVideoFrame<u8>> XcoderEncoder<F> {
             }
 
             frame.sei_total_len = 0;
-            if let Some(sei) = f.timecode().and_then(|tc| Self::sei_nalu(&self.config, tc)) {
+            if let Some(sei) = f.timecode().and_then(|tc| self.config.sei_nalu(tc)) {
                 // There's no way with this version of libxcoder to make the encoder write our
```
raven-txm, line 313: Looking forward to the next version, yep

raven-txm (dismissed):
Given that there's a lot of technical detail here I'm not totally conversant on, this looks good to me overall.

I like the refactors/improvements to code structure overall. I'll be exercising this code during testing of the ansible updates (incl. new NI firmware) on the northamerica hosts + temp logger and the like.

xcoder/xcoder-logan/src/encoder/xcoder_259.rs
ccbrown: I'm glad you commented on this, because I meant to remove it! It's not needed anymore since I refactored the way SEI NALUs are inserted.

ccbrown:
> Given that there's a lot of technical detail here I'm not totally conversant on, this looks good to me overall.
> 
> I like the refactors/improvements to code structure overall. I'll be exercising this code during testing of the ansible updates (incl. new NI firmware) on the northamerica hosts + temp logger and the like.

Two things to keep in mind:

* The current AVS won't work with the new firmware at all - the changes in this PR are strictly necessary.
* I couldn't get "cold" upgrades to the new firmware to work (see email thread), but forcing a "warm" upgrade via the `-w` flag to the upgrade script worked fine for me. You may want to do the same. If nothing else, it's easier because warm upgrades don't require restarts.

xcoder/xcoder-logan/src/encoder/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+    {
+        let mut nalus: Vec<&[u8]> = nalus.into_iter().collect();
+
+        // Insert SEI NALU if necessary.
+        let sei = timecode.and_then(|tc| self.sei_nalu(tc));
+        if let Some(sei) = &sei {
+            for (i, nalu) in nalus.iter().enumerate() {
+                if !nalu.is_empty() {
+                    let nal_unit_type = nalu[0] & h264::NAL_UNIT_TYPE_MASK;
+                    if nal_unit_type <= h264::NAL_UNIT_TYPE_CODED_SLICE_OF_IDR_PICTURE {
```
epingel2, line 118: So If I understand this correctly, this creates an sei_nalu from a timecode if given and then inserts that SEI before every IDR nalu. What was the previous logic? I don't see exactly where anything similar would have been done.

epingel2: Never mind. I was looking in the wrong repo entirely

epingel2 (approved):

xcoder/xcoder-logan/src/encoder/xcoder_259.rs
raven-txm: :D

raven-txm:
> The current AVS won't work with the new firmware at all - the changes in this PR are strictly necessary.

yes, I'm aware of this - the symbol changes alone pretty much make it not-backwards-compatible. I like what I'm seeing here and have no problem moving forward so I can test the firmware on the noram hosts soon with an eye to deploy next Tuesday.

I _like_ not having to do restarts during updates so much after last time....

raven-txm (approved):

PR #134, opened 2022-10-03
xilinx large packet segfault protection

ccbrown:
If you give the Xilinx decoder a packet with even 1 byte more than the number frame pixels, it either segfaults or errors out with "xma_plg_buffer_write failed. Can not write past end of buffer".

This adds a check to make sure it doesn't segfault.

I hate this change because it means any processed video absolutely _must_ have a certain compression ratio. But it's an improvement over segfaulting I guess.

### DIFF
--- a/xilinx/src/xlnx_decoder.rs
+++ b/xilinx/src/xlnx_decoder.rs
@@ -1,6 +1,6 @@
 use crate::sys::*;
 use crate::xlnx_dec_utils::*;
-use crate::XlnxError;
+use crate::{XlnxError, XlnxErrorType};
 use simple_error::{bail, SimpleError};
 
 pub struct XlnxDecoder {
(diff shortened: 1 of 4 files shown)


xilinx/src/xlnx_decoder.rs
```diff
@@ -49,6 +49,13 @@ impl XlnxDecoder {
     /// @buf: buffer of input data.
     /// @size: size of input data.
     pub fn xlnx_dec_send_pkt(&mut self, buf: &mut [u8], pts: i32) -> Result<(), XlnxError> {
+        if buf.len() > (self.frame_props.width * self.frame_props.height * self.frame_props.bits_per_pixel / 8) as usize {
```
scottlamb, line 52: Can we plumb this through as a permanent error rather than a transient one?

ccbrown: I added an `ErrorInvalid` variant to `XlnxErrorType` and am now using it here. I believe Xilinx only uses this error type when the input is invalid, so I think it's fitting for this purpose.

@epingel2 what do you think?

epingel2:
It would be better not to have this limitation, But as it is, preventing the program from needlessly segfaulting and instead providing useful error output is probably more useful to us than letting it segfault.

raven-txm (approved):
I'm not a huge fan. I dislike segfaults more.

PR #136, opened 2022-10-11
fix srt sockaddr conversion for rust 1.64

ccbrown:
Rust 1.64 changed the layout of `SocketAddr`: https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html

This makes our code not rely on that layout.


ccbrown:
Note the newer Logan device is currently offline so the xcoder-logan tests won't run (I'll bring it back up tomorrow.).

lzuosym (approved):

PR #137, opened 2022-10-14
fix forced key frames for netint logan 3.1.0 encoders

ccbrown:
NETINT made a sneaky change: In the 3.1.0 API, unlike the 2.X API, you can't use the "IDR" pict type to force a key frame. You have to use "FORCE_IDR". This fixes it and adds a test.


lzuosym (approved):

epingel2 (approved):
Looks good to me. I like the construction of the test video here for the test.

PR #138, opened 2022-10-14
Easier debugging

mtkopone:
Motivation: Capability to debug video issues in Statim.

* Allow setting the log level of the srt library from outside
* Build srt with ENABLE_DEBUG when not in release mode

### DIFF
--- a/srt/src/lib.rs
+++ b/srt/src/lib.rs
@@ -1,7 +1,7 @@
 #[macro_use]
 extern crate lazy_static;
 
-use srt_sys as sys;
+pub use srt_sys as sys;
 use std::{
     ffi::CStr,
     fmt,
@@ -446,10 +446,6 @@ pub struct ConnectOptions {
     pub too_late_packet_drop: Option<bool>,
     pub receive_buffer_size: Option<i32>,
     pub send_buffer_size: Option<i32>,
-}
-
-pub unsafe fn set_log_level(level: i32) {
-    sys::srt_setloglevel(level)
 }
 
 impl Stream {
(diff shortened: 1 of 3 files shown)


srt/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 #[macro_use]
 extern crate lazy_static;
 
-use srt_sys as sys;
+pub use srt_sys as sys;
```
ccbrown, line 4: I changed this up - the `srt` crate was meant to be a safer higher level abstraction, so I'd rather not just expose unsafe functions directly. But with this change, you can just call `srt::sys:: srt_setloglevel`.

srt/srt-sys/build.rs
```diff
@@ -1,9 +1,11 @@
 fn main() {
     let target_os = std::env::var("CARGO_CFG_TARGET_OS").expect("CARGO_CFG_TARGET_OS must be set");
     let target_arch = std::env::var("CARGO_CFG_TARGET_ARCH").expect("CARGO_CFG_TARGET_ARCH must be set");
+    let profile: String = std::env::var("PROFILE").expect("PROFILE must be set");
 
     let mut build = cmake::Config::new("vendor/srt-1.4.4");
     build.define("ENABLE_APPS", "OFF").define("ENABLE_SHARED", "OFF");
+    build.define("ENABLE_DEBUG", if profile == "release" { "OFF" } else { "ON" });
```
ccbrown, line 8: Have you confirmed that the debug build doesn't already have debug symbols? I'm pretty sure the `cmake` library does this already.

mtkopone: Looks like it gets properly set via probably CMAKE_BUILD_TYPE. So no need for this, reverted.

ccbrown (approved):

PR #141, opened 2022-10-17
Ensure decoder xvbm buffer pools are freed before xma_frames

epingel2:
Some applications experienced a seg fault if the decoder was dropped due to an abnormal exit. This was caused by the xframe being freed while leaving the xvbm buffer pool entries hanging. The xma decoder plugin would then attempt to free the entries when the application terminated causing a seg fault. This ensures that the buffer pool entries are freed before the xframe and not left hanging, so the decoder plugin will not attempt to free them during cleanup.

### DIFF
--- a/xilinx/src/xlnx_decoder.rs
+++ b/xilinx/src/xlnx_decoder.rs
@@ -133,6 +133,12 @@ impl Drop for XlnxDecoder {
                 xma_dec_session_destroy(self.dec_session);
             }
             if !self.out_frame.is_null() {
+                // when the decoder is dropped ensure that the xvbm_buffer_pool_entry has been freed
+                // there is no problem if xvbm_buffer_pool_entry_free is called more than once for
+                // an entry as long as the xframe exists
+                let handle: XvbmBufferHandle = (*self.out_frame).data[0].buffer;
+                xvbm_buffer_pool_entry_free(handle);
+
                 xma_frame_free(self.out_frame);
             }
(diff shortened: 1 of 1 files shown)


xilinx/src/xlnx_decoder.rs
```diff
@@ -133,6 +133,12 @@ impl Drop for XlnxDecoder {
                 xma_dec_session_destroy(self.dec_session);
             }
             if !self.out_frame.is_null() {
+                // when the decoder is dropped ensure that the xvbm_buffer_pool_entry has been freed
+                // there is no problem if xvbm_buffer_pool_entry_free is called more than once for
+                // an entry as long as the xframe exists
```
lzuosym, line 138: what is `xframe`? do you mean `out_frame`?

epingel2: the xframe is the data type of out_frame. I suppose I should probably rename it.

lzuosym (approved):

ccbrown:
> This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on April 1st, 2023. For more details, see https://github.com/actions/runner-images/issues/6002

Bah.

ccbrown (approved):

ccbrown:
We should be able to re-run the build for this in 36 minutes after the brownout ends.

PR #142, opened 2022-10-25
[AVS-583] interaving muxer causes the pts out of sync

lzuosym:
There is no issue for video only streams or the pts is less than `0x1FFFFFFFF`. However if the stream contains audio and video and we start to use deterministic pts,  `InterleavingMuxer` failed to generate the desired results. Check the unit test. I believe this is the actual root cause for the out of sync issue. I am working on the fix, but don't have a solution yet.

@ccbrown please also take a look.


lzuosym:
Seems like we can truncate the pts by `pst & 0x1FFFFFFFF` can fix the problem

mpeg2/src/interleaving_muxer.rs
```diff
@@ -596,4 +598,38 @@ mod test {
             ]
         );
     }
+
+    #[test]
+    fn test_audio_video_pts_from_wall_clock() {
```
ccbrown, line 603: Naming suggestion since it looks like the crux of the issue is that the input presentation times were higher than expected, and not necessarily that they were derived from wall clock times:

```suggestion
    fn test_oversized_presentation_times() {
```

ccbrown (commented):
Good find! Masking the input timestamps does seem like a good fix to me.

ccbrown (approved):
Lgtm, but add a second reviewer. 👍

epingel2 (approved):

PR #144, opened 2022-10-28
Add thread safe reference counter to srt api

epingel2:
Ensure that srt_startup and srt_cleanup are called in the correct order by using a reference counter. This addresses a bug where the API would get in a bad state in multithreaded applications when stopping and starting srt connections because srt_startup would be called on the new connection before srt_cleanup had been called on the previous one.

### DIFF
--- a/srt/src/lib.rs
+++ b/srt/src/lib.rs
@@ -10,7 +10,7 @@ use std::{
     net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6, ToSocketAddrs},
     pin::Pin,
     str,
-    sync::{Arc, Mutex, Weak},
+    sync::{Arc, Mutex, atomic::AtomicUsize},
 };
 
 #[cfg(feature = "async")]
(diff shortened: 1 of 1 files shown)


srt/src/lib.rs
```diff
@@ -109,33 +115,34 @@ impl Api {
...
-            None => {
-                unsafe {
-                    check_code("srt_startup", sys::srt_startup())?;
-                }
-                let new_api = Arc::new(Self::default());
-                *api = Some(Arc::downgrade(&new_api));
-                Ok(new_api)
+        let api_state = GLOBAL_API_STATE.lock().unwrap();
+        api_state.ref_count.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+        if api_state.ref_count.load(std::sync::atomic::Ordering::SeqCst) == 1 {
```
ccbrown, line 120: It'd be better to use the return value of `fetch_add` instead of doing another load, but...

Since we always hold the `api_state` mutex when we access `ref_count`, I don't think `ref_count` needs to be atomic at all. I think it can just be a plain `usize`.

epingel2: fetch_add returns the previous value which is annoying to me because it seems unintuitive. That's why I decided to fetch_add and then load.

epingel2: usizes aren't thread-safe by default though right?

ccbrown: They are if they're guarded by a mutex. And Rust wouldn't let you do anything with them that's not safe anyway.

epingel2: Yeah you're right that is all protected by that mutex

epingel2: Actually, because the API state is in an ARC within the global API state mutex, In order to mutate a ref_count I would have to either use a second `Mutex`, a `RwLock`, or just keep using the atomic types. It seems like the api state within the mutex should probably be in an ARC, so I think it might be the easiest option to keep the AtomicUsize

ccbrown (approved):

srt/src/lib.rs
ccbrown: Ah, I see. Yeah, this does seem easiest. Lgtm!

lzuosym (approved):
LGTM.

PR #149, opened 2022-12-27
fix compilation errors for bindgen 0.63.0

lzuosym:
Replace the deprecated methods in old bindgen


ccbrown (approved):
I guess we've put this off long enough. Hopefully this doesn't cause too much pain updating downstream crates.

PR #150, opened 2023-01-20
[AVS-681] fix HEVC encoding in high bitrate

lzuosym:
Before the HEVC encoding using Quadra hardware would either fail with insufficient resource error for bitrate>80mbps or result in unexpected transcoded bitrate for 70mbps<bitrate<80mbps. This will allow us to encode at a higher bitrate (maximum 800mbps) using Quadra hardware.

### DIFF
--- a/xcoder/xcoder-quadra/src/encoder.rs
+++ b/xcoder/xcoder-quadra/src/encoder.rs
@@ -204,7 +204,7 @@ impl<F> XcoderEncoder<F> {
                         // This specifies the size of the VBV (CPB) buffer in milliseconds.
                         // The range of supported values of `vbvBufferSize` is 0, and 10 to 3000. For example 3000 should be set for 3 seconds.
                         // Given a particular bitrate value, the maximum value of vbvBufferSize without leading to insufficient resource error is not documented.
-                        // For each bitrate range, we maximize vbvBufferSize value that does not exceed the Quadra encoder's processing capacity by trail and error.
+                        // For each bitrate range, we maximize vbvBufferSize value that does not exceed the Quadra encoder's processing capacity by trial and error.
                         cfg_enc_params.rc.vbv_buffer_size = match bitrate {
                             0..=80_000_000 => 3000,
                             80_000_001..=120_000_000 => 2000,
--- a/xcoder/xcoder-quadra/src/encoder.rs
+++ b/xcoder/xcoder-quadra/src/encoder.rs
@@ -203,8 +203,8 @@ impl<F> XcoderEncoder<F> {
                     if let Some(bitrate) = config.bitrate {
                         // This specifies the size of the VBV (CPB) buffer in milliseconds.
                         // The range of supported values of `vbvBufferSize` is 0, and 10 to 3000. For example 3000 should be set for 3 seconds.
-                        // The relationship between bitrate and vbvBufferSize is non-linear but unknown.
-                        // Given each bitrate range, we maximize the vbvBufferSize value while not exceeding the Quadra encoder's processing capacity.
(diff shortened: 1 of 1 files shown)


xcoder/xcoder-quadra/src/encoder.rs
```diff
@@ -200,6 +200,10 @@ impl<F> XcoderEncoder<F> {
                             XcoderH265Profile::Main10 => 2,
                         };
                     }
+                    // Rate control needs to be changed to VBR mode to encode at a higher bitrate (>60mbps).
```
scottlamb, line 203: Is this comment accurate? My understanding (which could be wrong) is that the VBV/CPB concept (at most n milliseconds of video can be buffered) still applies in VBR (variable bitrate) mode. And we switch from CBR (constant bitrate) and VBR by setting `cfg_enc_params.rc.crf` to something other than -1 (crf = Constant Rate Factor, as opposed to Constant Bitrate). `IntegrationProgrammingGuideQuadra_V4.0.pdf` says:

> When a value for `crf` is specified, it enables Constant Rate Factor mode. Lower values mean higher quality and higher values mean lower quality. CRF is a constant quality mode as opposed to normal rate control which is a constant bitrate (CBR). Note that unlike fixed QP operation, the QP during CRF can change from frame to frame. CRF sets QP in the range of {crf value-3 ~ crf value+3}, as well as minQp <= QP <= maxQp. When enabled, `rcEnable` must be 0 and `lookahead` must be enabled (lookaheadDepth=4~40).

How did you discover this option helped—trial and error? email from NETINT? if the latter, I'd be interested to see it.

lzuosym: The comment may not be accurate. This change is suggested by the NETINT support team. Please look at the comments in ticket AVS-681.

ccbrown: Here's the full email:

Screenshot 2023-01-20 at 14 43 31

Maybe there are other, better rate control settings we could use to get higher bitrates as well?

epingel2 (approved):
I think that we don't need to expose this to the caller. I believe we had a chat about this the other day and decided to just make the determination on the server side.

xcoder/xcoder-quadra/src/encoder.rs
lzuosym: Without `vbvBufferSize=0`, I found that the encoded video couldn't have the desired bitrate if the bitrate param passed to the encoder is larger than 65mbps. For example, if the bitrate is set to 80mbps, the actual bitrate in the encoded video is only 17mbps.

Maybe we can set `vbv_buffer_size=0` for all HEVC encoding?

scottlamb (approved):

xcoder/xcoder-quadra/src/encoder.rs
```diff
@@ -200,6 +200,18 @@ impl<F> XcoderEncoder<F> {
                             XcoderH265Profile::Main10 => 2,
                         };
                     }
+                    if let Some(bitrate) = config.bitrate {
+                        // This specifies the size of the VBV (CPB) buffer in milliseconds.
+                        // The range of supported values of `vbvBufferSize` is 0, and 10 to 3000. For example 3000 should be set for 3 seconds.
+                        // Given a particular bitrate value, the maximum value of vbvBufferSize without leading to insufficient resource error is not documented.
+                        // For each bitrate range, we maximize vbvBufferSize value that does not exceed the Quadra encoder's processing capacity by trail and error.
```
ccbrown, line 207:
```suggestion
                        // For each bitrate range, we maximize vbvBufferSize value that does not exceed the Quadra encoder's processing capacity by trial and error.
```

ccbrown (approved):

scottlamb (approved):

PR #152, opened 2023-01-25
[AVS-681] improve frame rate guess

lzuosym:
Before some 24fps videos with jitter timestamps are misclassified as 29.97fps videos. This improves the PTSAnalyzer's `guess_frame_rate` method.

1. Increased the number of frame samples for the guess to 26.
2. Fixed the above issue.

### DIFF
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -590,7 +590,7 @@ impl PTSAnalyzer {
     /// have jitter, e.g. due to being set to wall-clock times, the guess may be off. For those
     /// cases, it has a bias towards returning 23.98, 29.97 or 59.94.
     pub fn guess_frame_rate(&self) -> Option<f64> {
-        const MIN_TIMESTAMP_COUNT: usize = 26;
+        const MIN_TIMESTAMP_COUNT: usize = 16;
         const MAX_B_FRAMES: usize = 3;
 
         if self.timestamps.len() < MIN_TIMESTAMP_COUNT {
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -588,9 +588,9 @@ impl PTSAnalyzer {
     /// Makes a guess at a video's frame rate. This should really only be used as a last resort. If
     /// the presentation timestamps were set precisely it should be accurate, but if the timestamps
     /// have jitter, e.g. due to being set to wall-clock times, the guess may be off. For those
-    /// cases, it has a bias towards returning 23.98, 29.97 or 59.94.
+    /// cases, it has a bias towards returning 29.97 or 59.94.
     pub fn guess_frame_rate(&self) -> Option<f64> {
-        const MIN_TIMESTAMP_COUNT: usize = 16;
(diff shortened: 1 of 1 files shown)


mpegts-segmenter/src/analyzer.rs
```diff
@@ -942,4 +944,31 @@ mod test {
...
+            analyzer.flush().unwrap();
+        }
+
+        assert!(analyzer.has_video());
+        assert_eq!(
+            analyzer.streams()[0],
+            StreamInfo::Video {
+                width: 3840,
+                height: 2160,
+                frame_rate: 23.98,
```
ccbrown, line 966: Where does 23.98 come from? Is this what the camera outputs?

lzuosym: One is from [`BUF @ DET All-29 / 8K High 50`](https://dev.av.txm.sh/#/av-streams/CMFBv3IhUW5oBeV6RmcY6M4Nwlh-) in dev environment, which was ingested from a recording according to Ed. . Another example is `Chelsea CV Data Gathering 4K Replay Stream`, we discussed it in [av Slack channel](https://tempus-ex.slack.com/archives/C02M73ZAVCJ/p1673375934701249).  

Using netint Quadra for the HEVC transcoding job leads to incorrect fps video results.

ccbrown: In that case, I don't think there's anything special about the number 23.98 except that FFmpeg guesses tend to favor it. Our actual frame rate is just somewhere around ~24 FPS depending on how frames are dropped.

What happens if the bias towards 23.98 is removed? How close is the frame rate to 24 FPS?

lzuosym: The fps from `avg_delta` is ~25.71, `max_delta=4500`, `min_delta=2999`. Since `max_delta - min_delta` >> 5, the resulting guessed fps is 29.97 before this change.

Do you mean change it to ?
```diff
-            if (fps - 23.98).abs() < 3.0 {
-                return Some(23.98);
-            } else if (fps - 29.97).abs() < 3.0 {
+            if (fps - 29.97).abs() < 3.0 {
                 return Some(29.97);
```

And increasing `MIN_TIMESTAMP_COUNT` does improve the accuracy. I can try to set it to 20 or more .

scottlamb (commented):
Why is this guess important or necessary at all? I think that AVS-700 (for NETINT) and AVS-687 (for Xilinx) are necessary and sufficient to have transcodes reliably produce the right pts output.

lzuosym:
> Why is this guess important or necessary at all? I think that AVS-700 (for NETINT) and AVS-687 (for Xilinx) are necessary and sufficient to have transcodes reliably produce the right pts output.

You are right! From my testing,  NETINT Quadra transcoding precisely preserves the PTS in the encoded video. Seems like nothing need to be done for AVS-700.  And we don't need this change for netint to work properly, but for segments with jitter timestamp, the guessed framerate will be eventually persisted into DB and shown in av-frontend UI. It's kind of misleading if it's far away from the actual FPS.

ccbrown (approved):

epingel2 (approved):

PR #153, opened 2023-03-03
fix SRT async write

lzuosym:
This fixes SRT AsyncStream `poll_write`, so that it can accept data larger than SRT allowed payload size in UDP packets.

### DIFF
--- a/srt/src/async_lib.rs
+++ b/srt/src/async_lib.rs
@@ -229,7 +229,13 @@ impl AsyncWrite for AsyncStream {
             }
         }
         let sock = self.socket.raw();
-        let data_size = self.max_send_payload_size.min(src.len());
+        let data_size = if self.max_send_payload_size == 0 {
+            // When set to 0, there's no limit for a single sending call.
+            // https://github.com/Haivision/srt/blob/master/docs/API/API-socket-options.md#SRTO_PAYLOADSIZE
+            src.len()
+        } else {
+            self.max_send_payload_size.min(src.len())
+        };
 
         match unsafe { sys::srt_send(sock, src.as_ptr() as *const sys::char, data_size as _) } {
             len if len >= 0 => Poll::Ready(Ok(len as _)),
--- a/srt/src/async_lib.rs
+++ b/srt/src/async_lib.rs
@@ -289,7 +289,7 @@ mod test {
         let mut server_conn = accept_result.unwrap().0;
         let mut client_conn = connect_result.unwrap();
         assert_eq!(client_conn.id(), None);
-        assert_eq!(client_conn.payload_size(), 1316);
+        assert_eq!(client_conn.payload_size, 1316);
 
         let mut buf = [0; 1316];
         for i in 0..5 {
--- a/srt/src/async_lib.rs
+++ b/srt/src/async_lib.rs
@@ -125,7 +125,9 @@ impl AsyncStream {
     fn new(id: Option<String>, socket: Socket) -> Result<Self> {
         socket.set(sys::SRT_SOCKOPT_SRTO_SNDSYN, false)?;
         socket.set(sys::SRT_SOCKOPT_SRTO_RCVSYN, false)?;
-        let max_send_payload_size = socket.get::<i32>(sys::SRT_SOCKOPT_SRTO_PAYLOADSIZE).unwrap_or(DEFAULT_SEND_PAYLOAD_SIZE as _) as _;
+        let max_send_payload_size = socket
+            .get::<i32>(sys::SRT_SOCKOPT_SRTO_PAYLOADSIZE)
+            .expect("SRT should have a default payload size if not set") as _;
         Ok(Self {
             epoll_reactor: socket.api.get_epoll_reactor()?,
             socket,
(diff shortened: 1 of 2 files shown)


srt/src/async_lib.rs
```diff
@@ -279,6 +289,7 @@ mod test {
         let mut server_conn = accept_result.unwrap().0;
         let mut client_conn = connect_result.unwrap();
         assert_eq!(client_conn.id(), None);
+        assert_eq!(client_conn.payload_size(), 1316);
```
scottlamb: Maybe grow the payload on 285 to exercise the changes?

srt/src/async_lib.rs
```diff
@@ -217,17 +220,24 @@ impl AsyncRead for AsyncStream {
...
-                self.epoll_reactor.wake_when_write_ready(&self.socket, cx.waker().clone());
-                return Poll::Pending;
-            }
-        }
         let sock = self.socket.raw();
-        match unsafe { sys::srt_send(sock, src.as_ptr() as *const sys::char, src.len() as _) } {
-            len if len >= 0 => Poll::Ready(Ok(len as _)),
-            _ => Poll::Ready(Err(new_io_error("srt_send"))),
+        let mut sent = 0;
+        while sent < src.len() {
```
scottlamb, line 225: Isn't this loop producing small writes any time `src.len() % self.payload_size != 0`?

I don't think we need to loop. `tokio::io::AsyncWrite::poll_write` (like `std::io::Write::write`) is allowed to write only part of the bytes. If it's wrapped in a buffered writer with a buffer size >= `self.payload_size`, then we'll only produce small writes on flush, which is as it should be.

[`std::io::Write::write`](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write) actually forbids looping:

> A call to write represents *at most one* attempt to write to any wrapped object.

scottlamb (changes requested):
I think we should match this in the synchronous write path used by the agent. Do you want to do that in the same PR or different?

https://github.com/sportsball-ai/av-rs/blob/e169574a6720928cc9ab67862a69975a8b450508/srt/src/lib.rs#L532

lzuosym:
> I think we should match this in the synchronous write path used by the agent. Do you want to do that in the same PR or different?
> 
> https://github.com/sportsball-ai/av-rs/blob/e169574a6720928cc9ab67862a69975a8b450508/srt/src/lib.rs#L532

I can do it here.

srt/src/async_lib.rs
lzuosym: It's actually wrong to use a loop to send data async. If `Pending` returns from inside the loop, runtime will re-write the all data from beginning.

scottlamb (approved):

srt/src/lib.rs
```diff
@@ -492,13 +501,15 @@ impl Stream {
             let (addr, len) = to_sockaddr(&addr);
             let socket = Socket::new()?;
             socket.set_connect_options(options)?;
+            let max_send_payload_size = options.max_send_payload_size.unwrap_or(DEFAULT_SEND_PAYLOAD_SIZE as _) as _;
```
ccbrown, line 504: Would it make sense if instead of doing this, we checked `SRT_SOCKOPT_SRTO_PAYLOADSIZE` after the call to `srt_connect`?

And should we really default to `DEFAULT_SEND_PAYLOAD_SIZE`? Why not default to the previous behavior of not having a limit? Maybe that depends on in what circumstances `SRT_SOCKOPT_SRTO_PAYLOADSIZE` doesn't have a value?

lzuosym: > Would it make sense if instead of doing this, we checked `SRT_SOCKOPT_SRTO_PAYLOADSIZE` after the call to `srt_connect`?

Agree, this can be moved after the call to `srt_connect`.

> And should we really default to `DEFAULT_SEND_PAYLOAD_SIZE`? Why not default to the previous behavior of not having a limit? Maybe that depends on in what circumstances `SRT_SOCKOPT_SRTO_PAYLOADSIZE` doesn't have a value?

Even if `SRT_SOCKOPT_SRTO_PAYLOADSIZE` is not set, `socket.get::<i32>(sys::SRT_SOCKOPT_SRTO_PAYLOADSIZE)` will always have some value (1316 default) . So yes, it's not necessary to define a `DEFAULT_SEND_PAYLOAD_SIZE` const.

srt/src/async_lib.rs
```diff
@@ -224,7 +229,9 @@ impl AsyncWrite for AsyncStream {
             }
         }
         let sock = self.socket.raw();
-        match unsafe { sys::srt_send(sock, src.as_ptr() as *const sys::char, src.len() as _) } {
+        let data_size = self.max_send_payload_size.min(src.len());
```
ccbrown, line 232: A `0` value for `SRTO_PAYLOADSIZE` has a special meaning:

> When set to 0, there's no limit for a single sending call.

– https://github.com/Haivision/srt/blob/master/docs/API/API-socket-options.md#SRTO_PAYLOADSIZE

So maybe the logic should be something like...

```suggestion
        let data_size = if self.max_send_payload_size != 0 { self.max_send_payload_size.min(src.len()) } else { src.len() };
```

?

ccbrown (approved):

PR #154, opened 2023-03-13
[AVS-741] fix segment creation for audio only streams

lzuosym:
For the audio only streams, we start a new segment if the stream elapsed time is larger than the configured segment duration.


mpegts-segmenter/src/segmenter.rs
```diff
@@ -143,6 +143,9 @@ impl<S: SegmentStorage> Segmenter<S> {
                                                 }
                                             }
                                         }
+                                        Some(analyzer::Stream::ADTSAudio { .. }) => {
+                                            is_keyframe = elapsed_seconds < -1.0 || elapsed_seconds >= self.config.min_segment_duration.as_secs_f64().max(3.0);
```
ccbrown, line 147: This could potentially impact audio+video streams too, right?

Could we make this logic only apply for streams that don't have video?

lzuosym: I think for audio+video streams, if the mpegts packet is an audio packet, the following condition is false, so `should_start_new_segment` is not determined by the audio in this case.

```rust
      if (self.analyzer.is_video(p.packet_id) || !self.analyzer.has_video())
         && (elapsed_seconds < -1.0 || elapsed_seconds >= self.config.min_segment_duration.as_secs_f64())
```

ccbrown: Ah, I think you're right. Maybe a comment is warranted this:

```suggestion
                                            // note this is only reachable if this is an audio-only stream
                                            is_keyframe = elapsed_seconds < -1.0 || elapsed_seconds >= self.config.min_segment_duration.as_secs_f64().max(3.0);
```

epingel2: This was my question too earlier in standup @lzuosym Chris phrased it better.

epingel2 (approved):

scottlamb:
Err, this is embarrassing, I meant to do the squash and merge operation on my own PR, which was directly below in my inbox. The danger of doing things from my phone. That said, I do think this PR is ready to go...

lzuosym:
Never mind! As long as the PR is ready to go, it doesn't matter who clicks the merge button.

PR #155, opened 2023-05-18
[AVS-788] extract cropping info from mpeg-ts

lzuosym:
This extracts the cropping information from the MPEG-TS stream and populates the `SegmentInfo` struct with it, allowing for the further utilization.


mpegts-segmenter/src/segmenter.rs
```diff
@@ -7,12 +7,16 @@ use mpeg2::{
...
+pub const CROPPING_SIZE_IN_BYTES: usize = 156;
+pub const CROPPING_FLAG: [u8; 4] = [b't', b'x', b'm', b'0'];
+
 struct CurrentSegment<S: AsyncWrite + Unpin> {
     segment: S,
     pcr: u64,
     pts: Option<Duration>,
     bytes_written: usize,
     temi_timeline_descriptor: Option<TEMITimelineDescriptor>,
+    cropping: Vec<Vec<u8>>,
```
scottlamb, line 19: Can we just collect all private data without checking it's in a particular format? Then we confine all the knowledge of that format to the caller.

lzuosym: Yeah, it makes sense!

scottlamb (approved):

PR #156, opened 2023-05-18
lock rust-ffmpeg-sys version to fix the CI build

lzuosym:
Recent update in `rust-ffmpeg-sys` breaks the CI build, this locks the `rust-ffmpeg-sys` to a specific version that worked before. `Cargo.lock` file is also checked in to make the CI build reproducible. Otherwise the CI build would fail due to the update of the `time` dependency which requires the rust version 1.65.0, which we are currently building the code using 1.63.0.


scottlamb (approved):
(I'm assuming you'll update the pr description on merge)

scottlamb (approved):

PR #158, opened 2023-05-22
bundle video pts and private data together

lzuosym:


mpegts-segmenter/src/analyzer.rs
```diff
@@ -13,6 +13,12 @@ pub struct Timecode {
     pub frames: u8,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct VideoMetadata {
+    pub pts: u64,
```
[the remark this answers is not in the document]
lzuosym: changed it to relative pts!

mpegts-segmenter/src/analyzer.rs
```diff
@@ -46,11 +53,27 @@ pub enum Stream {
...
     },
     Other(u8),
 }
 
+fn convert_to_relative_pts(video_metadata: &[VideoMetadata]) -> Vec<VideoMetadata> {
+    const PTS_MOD: u64 = 1 << 33;
+    if video_metadata.is_empty() {
+        vec![]
+    } else {
+        let pts0 = video_metadata[0].pts;
```
scottlamb, line 66: can we do this based on the first packet in the segment, even if it doesn't have metadata? I'm trying to make this as robustly accurate as possible.

lzuosym: done!

scottlamb (approved):

PR #159, opened 2023-05-25
[AVS-861] reset stream metadata for each new segment

lzuosym:


scottlamb (changes requested):
Needs a test.

mpegts-segmenter/src/segmenter.rs
```diff
@@ -746,20 +748,23 @@ mod test {
...
-                    .all(|data| data.private_data.len() == 156 && data.private_data[..4] == [b't', b'x', b'm', b'0']));
-                assert_eq!(video_metadata.len(), 72);
-                let ptses = video_metadata.iter().map(|data| data.pts).collect::<Vec<u64>>();
-                assert_eq!(ptses.len(), 72);
-                for i in 1..ptses.len() {
-                    assert!(ptses[i] > ptses[i - 1]);
+        assert_eq!(segments.len(), 3);
+        let expected_video_metadata_len = [72, 72, 72];
+        for i in 0..3 {
+            match &segments[0].1.streams[0] {
```
scottlamb, line 754: `segments[0]`?

If I comment out the `self.analyzer.reset_stream_metadata();` on line 213, the test still passes, so it's not testing the right thing.

lzuosym: fixed it.

mpegts-segmenter/src/segmenter.rs
```diff
@@ -746,20 +748,23 @@ mod test {
...
+                StreamInfo::Video { video_metadata, .. } => {
+                    assert_eq!(video_metadata.len(), expected_video_metadata_len[i]);
+                    assert!(video_metadata
+                        .iter()
+                        .all(|data| data.private_data.len() == 80 && data.private_data[..4] == [b't', b'x', b'm', b'0']));
+
+                    let ptses = video_metadata.iter().map(|data| data.pts).collect::<Vec<u64>>();
+                    for i in 1..ptses.len() {
+                        assert!(ptses[i] > ptses[i - 1]);
```
scottlamb, line 763: Every frame in this file has private data. I'd also do `assert_eq!(ptses[0], 0);` and it fails. I don't think the pts reference is being reset correctly.

mpegts-segmenter/src/segmenter.rs
```diff
@@ -746,20 +748,23 @@ mod test {
...
-                    .iter()
-                    .all(|data| data.private_data.len() == 156 && data.private_data[..4] == [b't', b'x', b'm', b'0']));
-                assert_eq!(video_metadata.len(), 72);
-                let ptses = video_metadata.iter().map(|data| data.pts).collect::<Vec<u64>>();
-                assert_eq!(ptses.len(), 72);
-                for i in 1..ptses.len() {
-                    assert!(ptses[i] > ptses[i - 1]);
+        assert_eq!(segments.len(), 3);
+        let expected_video_metadata_len = [72, 72, 72];
+        for i in 0..3 {
```
scottlamb, line 753: btw, I'd just do `for seg in &segments` and then `match &seg.1.streams[0]`. More readable, harder to make the `[i]`->`[0]` mistake.

mpegts-segmenter/src/segmenter.rs
lzuosym: Good point!

mpegts-segmenter/src/segmenter.rs
lzuosym: done

scottlamb (approved):

PR #161, opened 2023-06-21
[AVS-881] expose segment duration

lzuosym:
This enables the MPEG-TS segmenter to determine the duration of segments based on the frame's presentation time.


mpegts-segmenter/src/analyzer.rs
```diff
@@ -838,6 +900,7 @@ mod test {
                     timecode: None,
                     is_interlaced: false,
                     video_metadata: vec![],
+                    duration: Some(Duration::from_secs_f64(1.017827794)),
```
lzuosym, line 903: The duration of the test video clip is about 1s, which matches the duration calculated here. But I don't understand why  `frame_count`/`frame_rate`=0.51718 is only half of the actual duration.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -694,6 +737,22 @@ impl PTSAnalyzer {
 
         Some((fps * 100.0).round() / 100.0)
     }
+
+    pub fn segment_duration(&self, frame_rate: f64) -> Option<Duration> {
```
scottlamb, line 741: We need to do this by a simple delta of the pts that starts the next segment and the pts that ends this segment. Using `frame_rate` is wrong.

scottlamb (changes requested):

mpegts-segmenter/src/analyzer.rs
scottlamb: > 1. For video clips that only fit into one segment, there is no next segment that can be used to calculate the delta. This particular special case shouldn't be a cause for concern except in unit tests.

Yeah, I think we should just update unit test data as needed to match our real usage.

> If several or more frames drop after the end of the first segment, the duration calculated using the delta approach includes the missing frames, which could results in more errors.

I agree this case is critical, but can you explain why you think it's correct to not include these in the duration? My understanding is: a frame's duration is defined as the time until the following frame's pts. If say frame 5/30 is dropped, frame 4's duration is therefore extended. If frame 30/30 is dropped, likewise frame 29/30's duration is extended in the player until the following segment's initial pts. And the whole segment's duration is unchanged—it still lasts until the following segment starts. Segment-level info used in manifests and things should still reflect that the segment lasts until the following segment, right?

lzuosym: > I agree this case is critical, but can you explain why you think it's correct to not include these in the duration? My understanding is: a frame's duration is defined as the time until the following frame's pts. If say frame 5/30 is dropped, frame 4's duration is therefore extended. If frame 30/30 is dropped, likewise frame 29/30's duration is extended in the player until the following segment's initial pts. And the whole segment's duration is unchanged—it still lasts until the following segment starts. Segment-level info used in manifests and things should still reflect that the segment lasts until the following segment, right?

That's interesting! I'm not familiar with the HLS manifest standard. If the segment duration in the HLS manifest is as you mentioned, then it would be appropriate to use the delta approach!

mpegts-segmenter/src/segmenter.rs
```diff
@@ -528,6 +551,16 @@ mod test {
                 11.423_221_999_999_999,
             ]
         );
+
+        let durations = segments.iter().flat_map(|(_, info)| info.duration).collect::<Vec<u64>>();
+        assert_eq!(
+            durations,
+            [
+                90000, 90000, 90000, 90000, 90000, 90000, 90000, 90000, 90000, 90000, 90000, 90000, 90000, 8588764592, 90000, 90000, 90000, 90000, 90000,
```
lzuosym, line 559: How to we handle the cases due to PCR to reset? The segment duration to be extremely large (8588764592), it it okay for the HLS manifests? Should we handle this case in a special manner? What are your suggestions or ideas on approaching it?

mpegts-segmenter/src/segmenter.rs
```diff
@@ -230,6 +237,16 @@ impl<S: SegmentStorage> Segmenter<S> {
                         }
                     }
                     segment.pts = pes_packet_header.and_then(|h| h.optional_header).and_then(|h| h.pts);
+                    if let Some((previous_segment, mut previous_segment_info)) = self.previous_segment.take().zip(self.previous_segment_info.take()) {
+                        if let Some((prev_pts, curr_pts)) = previous_segment.pts.zip(segment.pts) {
+                            let duration = (curr_pts + PTS_ROLLOVER_MOD - prev_pts) % PTS_ROLLOVER_MOD;
+                            // set the segment duration only if it's within a reasonable value.
+                            if duration < (self.config.min_segment_duration.as_secs_f64() * VIDEO_PTS_BASE as f64 * 3.0) as u64 {
```
lzuosym, line 244: We can raise an error at this point if the duration is an unreasonable value, but it results in the loss of two segments(current one and previous one). Additionally, the unit test `test_segmenter_restart` will fail due to this error. Alternatively, by setting the segment duration to `None`, the streaming can continue, and we can include an error message in the av-service log if it's necessary.

scottlamb: My feeling is that it's okay to discard both segments if the pts delta is crazy:

* We should make every effort on the agent side anyway to have correct, monotonically increasing timestamps by e.g. having a NTP-synced wall clock, taking care with our pts on Continuum input switch, etc., so that HA works as well as it can, so that sync between streams works well, etc.
* Not sure there's anything reasonable the player can do with a segment of unknown duration anyway. I really want to get rid of any temptation for the service, fusion feed, the player, etc. to add in heuristics and then have things just get off. The player folks need everything to be frame-accurate for AR, so I don't want to fudge it anywhere.
* Eventually, I'm tempted to switch from segments being stored as MPEG-TS to segments being stored as fMP4/CMAF. For those, we have to come up with a duration.

Strictly speaking, we don't *have* to drop the entire previous segment if the duration is crazy; we could basically go back to the previous video frame or something. But that probably requires an `SegmentStorage` interface change or something and doesn't seem work the complexity.

@epingel2 do you have opinions on this?

lzuosym: I don't know which one have more impact on the player, dropping segments or the duration of unknown segments. @maredudd what are your thoughts on this?

PR #162, opened 2023-06-21
use ffmpeg-sys-next, single version of bindgen

scottlamb:
One version is easier to keep track of and should reduce build times.

* replace ffmpeg-sys with ffmpeg-sys-next, which (among other things) uses bindgen 0.64. ffmpeg-sys was using a pre-0.60 version, so in some of our call sites, it hit build errors on Rust 1.70 due to https://github.com/rust-lang/rust-bindgen/issues/2083 https://github.com/rust-lang/rust-bindgen/blob/main/CHANGELOG.md#0600

* update cargo lock stuff to use the same version

* remove stale comments about older versions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 3
 
 [[package]]
-name = "aho-corasick"
-version = "1.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "67fc08ce920c31afb70f013dcce1bfc3a3195de6a228474e45e1f145b36f8d04"
-dependencies = [
- "memchr",
-]
-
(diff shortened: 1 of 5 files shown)


scottlamb:
Hold off on reviewing for a minute. I tried using this and hit a problem with xilinx. Gonna see if I can easily make the CI actually build that to verify the fix / avoid introducing such problems again.

scottlamb:
Better now. I gave up for now on making xilinx build in CI (the SDK wants older versions of apt deps than the `ubuntu-20.4` image wants to install, ugh), but I fixed the compilation error as tested externally.

epingel2 (approved):

PR #163, opened 2023-06-23
Add BGRA input support to video-toolbox and x264

epingel2:


macos/core-video/src/pixel_buffer.rs
```diff
@@ -13,6 +13,31 @@ pub struct PixelBufferPlane {
...
+    {
+        let mut ret = std::ptr::null_mut();
+        result(
+            sys::CVPixelBufferCreateWithBytes(
+                std::ptr::null(), // allocator
+                width as _,
+                height as _,
+                pixel_format_type,
+                data_ptr as _, // dataPtr
+                (width * 4) as usize,            // dataSize
```
scottlamb, line 30: This is dependent on `pixel_format_type`, right? If we ever want to use a pixel format where this isn't the right value, we'll trip over this hidden assumption. A couple ideas:

* make the caller specify it (if/for as long as we keep this interface `unsafe`, see other comment)
* use the `VideoEncoderInputFormat` enum instead, and have it produce both the `pixel_format_type` and the appropriate calculation for a given width (whether that's N bits per pixel, some enforced padding, or whatever).

macos/core-video/src/pixel_buffer.rs
```diff
@@ -13,6 +13,31 @@ pub struct PixelBufferPlane {
 }
 
 impl PixelBuffer {
+    /// Creates a new pixel buffer that references the bytes, without copying them.
+    ///
+    /// # Safety
+    /// This is unsafe because the caller must ensure that the given bytes out-live the PixelBuffer.
```
scottlamb, line 19: This follows the existing pattern, so I think it's fine for now, but fwiw I think it shouldn't be too hard to make this safe. Make it `PixelBuffer<'a>`, pass in a `&'a mut [u8]`, verify the buf length is consistent with other parameters, and keep the lifetime via a `PhantomData<&'a mut ()>` or some such. Likewise `PixelBufferPlane` would be `PixelBufferPlane<'a>` and have `pub data: &'a mut [u8]`. Maybe add a `TODO` if you agree?

macos/core-video/src/pixel_buffer.rs
epingel2: Yes. Thanks for pointing this out. I was looking at just this narrow use case, but this could really cause a difficult to find issue in the future

scottlamb (approved):

PR #164, opened 2023-06-29
[AVS-864] refactor mpegts-segmenter

lzuosym:
This refine the process of collecting information that  is used for create a new segment. 
1. Improved the process of reseting the stream statistics via a single loop.
2. Removed `streams_before_segment` from `Segmenter ` so that the `SegmentInfo` is solely based on the statistics from the creation of the last segment.
3. `pts` in `VideoMetadata` is set to the relative to the start of the segment without the need to be adjusted before persisting it.
4. `pes_packet_header` is reused for getting the segment pts and video metadata.


mpegts-segmenter/src/analyzer.rs
```diff
@@ -464,24 +463,12 @@ impl Analyzer {
             .collect()
     }
 
-    pub fn reset_stream_metadata(&mut self) {
+    pub fn reset_streams_data(&mut self) {
```
scottlamb, line 466: Would it make sense to combine this with `.streams()`, something like `.take_segment_info()`, to return info about the just-completed segments and set up for the next one at once?

Would it make sense to do the same with the data structures, e.g. within `Stream` to have the per-segment stuff in a separate `struct` we can reset all in one go?

mpegts-segmenter/src/analyzer.rs
```diff
@@ -334,21 +334,20 @@ impl Stream {
...
         if let Self::AVCVideo { timecode, .. } = self {
             *timecode = None;
         }
     }
 
+    pub fn reset_frame_counter(&mut self) {
+        match self {
+            Stream::AVCVideo { access_unit_counter, .. } => *access_unit_counter = h264::AccessUnitCounter::new(),
+            Stream::HEVCVideo { access_unit_counter, .. } => *access_unit_counter = h265::AccessUnitCounter::new(),
+            _ => {}
```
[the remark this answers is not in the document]
lzuosym: Good catch and thanks! 👍

mpegts-segmenter/src/analyzer.rs
lzuosym: > Would it make sense to combine this with `.streams()`, something like `.take_segment_info()`, to return info about the just-completed segments and set up for the next one at once?

I think the `.streams()` API is like an HTTP `GET` method and it doesn't reset `analyzer`, and it doesn't have to be called when new segment starts or ends.

> Would it make sense to do the same with the data structures, e.g. within `Stream` to have the per-segment stuff in a separate `struct` we can reset all in one go?

Same reason for `info` method in `Stream` struct, we don't reset anything in `.info()`.  I am not sure if understand this correctly, what do you mean "in a separate `struct`"?

mpegts-segmenter/src/segmenter.rs
```diff
@@ -170,45 +170,17 @@ impl<S: SegmentStorage> Segmenter<S> {
...
-                            }
-                        }
-                        pes_packet_header = Some(header);
-                    }
-                }
-            } else {
-                first_temi_timeline_descriptor = None;
-            }
-
+            let mut prev_segment = None;
```
epingel2, line 173: This can be refactored to be more readable. And then could be broken into other functions:
This removes the need for a "PreviousSegment" struct entirely. I think this write function is huge and can be broken up further into smaller single-responsibility functions. 
```
            let mut next_segment_start_pts = None;
            let mut pes_packet_header: Option<pes::PacketHeader> = None;
            if should_start_new_segment {
                if let Some(cur) = self.current_segment.take() {
                    let mut segment_info = SegmentInfo::compile(&cur, self.analyzer.streams());
                    
                    if self.analyzer.is_pes(p.packet_id) && p.payload_unit_start_indicator {
                        if let Some(payload) = p.payload.as_ref() {
                            pes_packet_header = Some(pes::PacketHeader::decode(payload)?.0);
                            next_segment_start_pts = pes_packet_header.as_ref().and_then(|h| h.optional_header.as_ref()).and_then(|h| h.pts);
                        }
                    }

                    if let Some((prev_pts, curr_pts)) = cur.pts.zip(next_segment_start_pts) {
                        let duration = (curr_pts + PTS_ROLLOVER_MOD - prev_pts) % PTS_ROLLOVER_MOD;
                        let max_reasonable_duration = (self.config.min_segment_duration.as_secs_f64() * VIDEO_PTS_BASE as f64 * 3.0) as u64;
                        // set the segment duration only if it's within a reasonable value.
                        if duration < max_reasonable_duration {
                            segment_info.duration = Some(duration);
                        } else {
                            return Err(Error::Other(
                                format!(
                                    "The segment duration {} exceeds maximum of {}; pts ranges from {} to {}.",
                                    duration, max_reasonable_duration, prev_pts, curr_pts,
                                )
                                .into(),
                            ));
                        }
                    }
                    self.storage.finalize_segment(cur.segment, segment_info).await?;
                }

                //Set up the new segment
                self.current_segment = Some(CurrentSegment {
                    segment: self.storage.new_segment().await?,
                    pcr: self.pcr.unwrap_or(0),
                    pts: next_segment_start_pts,
                    bytes_written: 0,
                    temi_timeline_descriptor: None,
                });

                self.analyzer.reset_streams_data();
            }

            if let Some(segment) = &mut self.current_segment {
                if let Some(af) = p.adaptation_field {
                    if segment.temi_timeline_descriptor.is_none() {
                        segment.temi_timeline_descriptor = af.temi_timeline_descriptors.into_iter().next();
                    }
                    if !af.private_data_bytes.is_empty() {
                        // error out if the private data in a MPEG-TS packet is not in the beginning of PES
                        if self.analyzer.is_pes(p.packet_id) && !p.payload_unit_start_indicator {
                            return Err(Error::Mpeg2Decode(mpeg2::DecodeError::new("private data is not in the beginning of PES.")));
                        }
                        if let Some(payload) = p.payload {
                            let header = pes_packet_header.unwrap_or(pes::PacketHeader::decode(&payload)?.0);
                            if let Some(pts) = header.optional_header.as_ref().and_then(|h| h.pts) {
                                match self.analyzer.stream(p.packet_id) {
                                    Some(analyzer::Stream::AVCVideo { video_metadata, .. }) | Some(analyzer::Stream::HEVCVideo { video_metadata, .. }) => {
                                        add_video_metadata(video_metadata, segment.pts, pts, af.private_data_bytes.into_owned());
                                    }
                                    _ => {}
                                }
                            }
                        }
                    }
                }

                segment.segment.write_all(buf).await?;
                segment.bytes_written += buf.len();
            }
        }
        Ok(())
    }
```

mpegts-segmenter/src/analyzer.rs
epingel2: Hmm. I don't think that the analyzer should have any concept of a "Segment". However, I also think that a function called "reset_streams_data" is rather vague since obviously there's lots of data we aren't resetting. We might be able to create a function within the segmenter that is responsible for calling streams() and also resetting the particular fields on each stream and containing that logic on that side.

epingel2 (changes requested):

mpegts-segmenter/src/analyzer.rs
lzuosym: > Hmm. I don't think that the analyzer should have any concept of a "Segment". However, I also think that a function called "reset_streams_data" is rather vague since obviously there's lots of data we aren't resetting. We might be able to create a function within the segmenter that is responsible for calling streams() and also resetting the particular fields on each stream and containing that logic on that side.

The data that need to be reset are all inside the `Analyzer`. I think having a reset function isolates the Analyzer internal operations from the Segmenter.  Yes,  lots of data we aren't resetting, the function name is confusing, maybe we can just call it `reset_for_new_segment`? 🤷

mpegts-segmenter/src/segmenter.rs
```diff
@@ -19,6 +19,12 @@ struct CurrentSegment<S: AsyncWrite + Unpin> {
     temi_timeline_descriptor: Option<TEMITimelineDescriptor>,
 }
 
+struct PrevSegment<S: AsyncWrite + Unpin> {
```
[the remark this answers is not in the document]
lzuosym: We need to calculate the duration of segment by `CurrentSegment.pts - PrevSegment.pts`. This requires the `CurrentSegment` to be available before calling method `storage.finalize_segment` on the `PrevSegment`. However the `SegmentInfo` for the `PrevSegment` needs to be collected, and the Analyzer needs to be reseted before analyzing the `CurrentSegment` data. Not sure if I answered you question.

mpegts-segmenter/src/segmenter.rs
lzuosym: There is one problem with this code. `self.storage.finalize_segment(cur.segment, segment_info).await?;` and `self.analyzer.reset_streams_data();` need to be called before starting analyzing the data for the new segment.
```                    if self.analyzer.is_pes(p.packet_id) && p.payload_unit_start_indicator {
                        if let Some(payload) = p.payload.as_ref() {
                            pes_packet_header = Some(pes::PacketHeader::decode(payload)?.0);
                            next_segment_start_pts = pes_packet_header.as_ref().and_then(|h| h.optional_header.as_ref()).and_then(|h| h.pts);
                        }
                    }
```

epingel2:
This never got to the point that it actually fufilled the desired requirements and should be looked at again some day, but we are not ready to move forward with it right now.

PR #167, opened 2023-08-25
more encoding flexibility

scottlamb:
* in video toolbox H.265 encoder, allow setting level and max bitrate. Note I originally would automatically cap the max bitrate based on the level, but this might have enough undesirable effects that it's worth being able to play with it in the caller. They can use `H265ProfileLevel::max_bitrate` if desired.

* represent the encoder dropping a frame (altering the trait for all encoders). Video toolbox apparently does this sometimes based on the level and max bitrate selected; I had to track down a confusing `OSStatus(0)` locally to end up here.

* add some context to errors on video encoder creation. It turns out the GitHub Actions CI machine doesn't support setting data rate limits, but my M1 laptop does.


macos/core-foundation/src/string.rs
```diff
@@ -0,0 +1,29 @@
+use super::*;
+
+#[repr(transparent)]
+pub struct StringRef(sys::CFStringRef);
+
+impl StringRef {
+    pub fn from_static(value: &'static str) -> Self {
+        Self(unsafe {
+            sys::CFStringCreateWithBytesNoCopy(
+                std::ptr::null_mut(),
```
lzuosym, line 10: same question `null()` should be good.

macos/core-foundation/src/array.rs
```diff
@@ -23,3 +23,19 @@ impl Array {
         }
     }
 }
+
+impl<T: CFType> From<&[T]> for Array {
+    fn from(value: &[T]) -> Self {
+        Self(unsafe { sys::CFArrayCreate(std::ptr::null_mut(), value.as_ptr() as _, value.len() as _, &sys::kCFTypeArrayCallBacks) })
```
lzuosym, line 29: I think it doesn't need to be `null_mut()`.

macos/video-toolbox/build.rs
```diff
@@ -20,6 +20,7 @@ fn main() {
             .allowlist_var("kVTVideoEncoderSpecification_.+")
             .allowlist_var("kVTCompressionPropertyKey_.+")
             .allowlist_var("kVTEncodeFrameOptionKey_.+")
+            .allowlist_var("kVTProfileLevel_.+")
```
lzuosym, line 23: do we have to expose variables with prefix `kVTProfileLevel_`?

macos/video-toolbox/src/video_encoder.rs
```diff
@@ -1,14 +1,109 @@
...
+            H265ProfileLevel::Main3_0 => "HEVC_Main_3_0",
+            H265ProfileLevel::Main3_1 => "HEVC_Main_3_1",
+            H265ProfileLevel::Main4_0 => "HEVC_Main_4_0",
+            H265ProfileLevel::Main4_1 => "HEVC_Main_4_1",
+            H265ProfileLevel::Main5_0 => "HEVC_Main_5_0",
+            H265ProfileLevel::Main5_1 => "HEVC_Main_5_1",
+            H265ProfileLevel::Main5_2 => "HEVC_Main_5_2",
+            H265ProfileLevel::Main6_0 => "HEVC_Main_6_0",
+            H265ProfileLevel::Main6_1 => "HEVC_Main_6_1",
+            H265ProfileLevel::Main6_2 => "HEVC_Main_6_2",
```
lzuosym, line 93: just curious where do these string values documented for each profile`"HEVC_Main_6_2", "HEVC_Main_6_1"`...?

lzuosym (changes requested):

macos/video-toolbox/build.rs
scottlamb: No. I was trying to use them, but the constants for specific HEVC levels don't exist anyway, so...removed.

macos/video-toolbox/src/video_encoder.rs
scottlamb: They're not. Added a comment about it.

lzuosym (approved):

macos/video-toolbox/build.rs
lzuosym: It hasn't been removed yet, but there's no harm in keeping it.

PR #168, opened 2023-08-29
more Video Toolbox knobs

scottlamb:
* expose a few more encoder properties, and move them into a separate struct that supports `Default` so the caller can use `..Default::default()` rather than listing them all.

* add a tool to list all encoders and their supported properties

* Core Foundation extensions to support the above


lzuosym (approved):
I can't see any issues. It looks good to me! 👍

PR #169, opened 2023-08-30
time creation using CMTimeMake function

lzuosym:
The `flags` is set to be `0` in `CMTime` variables, which seems to be invalid. 
This uses a Core Media API `CMTimeMake` to create variables of `CMTime` type.


lzuosym:
Close it as it breaks the API's compatibility.

PR #171, opened 2023-09-15
vendored openssl should disable openssl pkg-config

scottlamb:
I saw strange/inconsistent behavior when compiling a dependent project. It was picking up homebrew-installed openssl@3 and failing on the lack of `openssl/evp.h`. After a `cargo clean`, I can't seem to reproduce. Still, it doesn't make sense to use both vendored openssl and look for another openssl via pkg-config, so let's not do that.

I compiled both `srt` itself and a dependent project after `cargo clean` in a few configs:

* no change => success (huh)
* this change only => success
* `CMakeLists.txt` hack below only => fail
* this change + `CMakeLists.txt` hack => succeeds

The hack to check that the `USE_OPENSSL_PC` define is making it through:

```
diff --git a/srt/srt-sys/vendor/srt-1.4.4/CMakeLists.txt b/srt/srt-sys/vendor/srt-1.4.4/CMakeLists.txt
index 21478c3..357f2b5 100644
--- a/srt/srt-sys/vendor/srt-1.4.4/CMakeLists.txt
+++ b/srt/srt-sys/vendor/srt-1.4.4/CMakeLists.txt
@@ -339,7 +339,7 @@ if (ENABLE_ENCRYPTION)
                # Try using pkg-config method first if enabled,
                # fall back to find_package method otherwise
                if (USE_OPENSSL_PC)
-                       pkg_check_modules(SSL ${SSL_REQUIRED_MODULES})
+                       message(FATAL_ERROR "openssl pc should be disabled")
                endif()
                if (SSL_FOUND)
                        # We have some cases when pkg-config is improperly configured
```

### DIFF
--- a/srt/srt-sys/build.rs
+++ b/srt/srt-sys/build.rs
@@ -13,6 +13,7 @@ fn main() {
                 arch => arch,
             },
         );
+        build.define("USE_OPENSSL_PC", "OFF");
         build.define("OPENSSL_INCLUDE_DIR", format!("{}/vendor/openssl/include", env!("CARGO_MANIFEST_DIR")));
         build.define(
             "OPENSSL_CRYPTO_LIBRARY",


lzuosym (approved):
Good catch! 👍  
Without this change, I think it's always picking the system installed openssl using `pkg-config`. Variables `OPENSSL_*` are not even used.

```log
[srt-sys 0.1.0]
    Building [srt-sys 0.1.0] -- Checking for module 'openssl libcrypto'
    [srt-sys 0.1.0] --   Found openssl libcrypto, version 3.1.2;3.1.2
[srt-sys 0.1.0] -- SSL via pkg-config: -L /opt/homebrew/Cellar/openssl@3/3.1.2/lib -I /opt/homebrew/Cellar/openssl@3/3.1.2/include -l;ssl;crypto
[srt-sys 0.1.0] -- ENCRYPTION: ENABLED, using: openssl libcrypto
[srt-sys 0.1.0] CMake Warning:
[srt-sys 0.1.0]   Manually-specified variables were not used by the project:
[srt-sys 0.1.0]
[srt-sys 0.1.0]     CMAKE_ASM_COMPILER
[srt-sys 0.1.0]     CMAKE_ASM_FLAGS
[srt-sys 0.1.0]     OPENSSL_CRYPTO_LIBRARY
[srt-sys 0.1.0]     OPENSSL_INCLUDE_DIR
[srt-sys 0.1.0]     OPENSSL_SSL_LIBRARY
```

Here is the `cargo build --vvv` output with `USE_OPENSSL_PC=OFF`
```log
[srt-sys 0.1.0] -- Found OpenSSL: /Users/leon/dev/tmp/av-rs/srt/srt-sys/vendor/openssl/lib/macos-aarch64/libcrypto.a (found version "3.0.1")
[srt-sys 0.1.0] -- SSL_LIBRARIES: /Users/leon/dev/tmp/av-rs/srt/srt-sys/vendor/openssl/lib/macos-aarch64/libcrypto.a
[srt-sys 0.1.0] -- SSL_INCLUDE_DIRS: /Users/leon/dev/tmp/av-rs/srt/srt-sys/vendor/openssl/include
[srt-sys 0.1.0] -- SSL via find_package(OpenSSL): -I /Users/leon/dev/tmp/av-rs/srt/srt-sys/vendor/openssl/include -l;/Users/leon/dev/tmp/av-rs/srt/srt-sys/vendor/openssl/lib/macos-aarch64/libcrypto.a
[srt-sys 0.1.0] -- ENCRYPTION: ENABLED, using: openssl libcrypto
[srt-sys 0.1.0] -- SSL libraries: /Users/leon/dev/tmp/av-rs/srt/srt-sys/vendor/openssl/lib/macos-aarch64/libcrypto.a
[srt-sys 0.1.0] CMake Warning:
[srt-sys 0.1.0]   Manually-specified variables were not used by the project:
[srt-sys 0.1.0]
[srt-sys 0.1.0]     CMAKE_ASM_COMPILER
[srt-sys 0.1.0]     CMAKE_ASM_FLAGS
[srt-sys 0.1.0]
[srt-sys 0.1.0]

```

PR #172, opened 2023-09-16
some Video Toolbox session tracing

scottlamb:
It's occasionally helpful to have a final verification that the expected properties are being set. It's quite verbose, so only do it when the tracing level is cranked up all the way.


lzuosym (approved):
LGTM!

PR #176, opened 2023-11-20
[AVS-1020] Update to v2 allocation and reservation functions

epingel2:
This updates to the "v2" allocation and reservation functions and structures present in xilinx 3.0 These are the structures and functions used by xilinx's ffmpeg and some of their example code. They allow a reduction in some code duplication and a more accurate allocation and reservation of resources based both on device Id as well as reservation id

### DIFF
--- a/xilinx/src/xlnx_dec_utils.rs
+++ b/xilinx/src/xlnx_dec_utils.rs
@@ -16,7 +16,7 @@ pub struct XlnxDecBuffer<'a> {
 
 pub struct XlnxDecoderXrmCtx {
     pub xrm_reserve_id: Option<u64>,
-    pub device_id: Option<u64>,
+    pub device_id: Option<u32>,
     pub dec_load: i32,
     pub decode_res_in_use: bool,
     pub xrm_ctx: xrmContext,
@@ -67,13 +67,13 @@ pub fn xlnx_calc_dec_load(xrm_ctx: xrmContext, xma_dec_props: *mut XmaDecoderPro
     Ok(load)
 }
 
-fn xlnx_fill_dec_pool_props(cu_pool_prop: &mut xrmCuPoolPropertyV2, dec_load: i32, device_id: Option<u64>) -> Result<(), SimpleError> {
+fn xlnx_fill_dec_pool_props(cu_pool_prop: &mut xrmCuPoolPropertyV2, dec_load: i32, device_id: Option<u32>) -> Result<(), SimpleError> {
     cu_pool_prop.cuListNum = 1;
 
     let mut device_info = 0;
 
     if let Some(device_id) = device_id {
-        device_info = (device_id << XRM_DEVICE_INFO_DEVICE_INDEX_SHIFT)
+        device_info = (device_id << XRM_DEVICE_INFO_DEVICE_INDEX_SHIFT) as u64
             | ((XRM_DEVICE_INFO_CONSTRAINT_TYPE_HARDWARE_DEVICE_INDEX as u64) << XRM_DEVICE_INFO_CONSTRAINT_TYPE_SHIFT);
     }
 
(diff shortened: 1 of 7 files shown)


xilinx/src/xlnx_dec_utils.rs
```diff
@@ -67,105 +67,58 @@ pub fn xlnx_calc_dec_load(xrm_ctx: xrmContext, xma_dec_props: *mut XmaDecoderPro
     Ok(load)
 }
 
-fn xlnx_fill_dec_pool_props(cu_pool_prop: &mut xrmCuPoolProperty, dec_load: i32) -> Result<(), SimpleError> {
-    cu_pool_prop.cuListProp.sameDevice = true;
+fn xlnx_fill_dec_pool_props(cu_pool_prop: &mut xrmCuPoolPropertyV2, dec_load: i32, device_id: Option<u64>) -> Result<(), SimpleError> {
```
scottlamb, line 70: Maybe `Option<u32>`, given that the header says `#define XRM_DEVICE_INFO_DEVICE_INDEX_MASK 0xFFFFFFFF`? Giving a `device_id` larger than a `u32` would overflow into the other parts of the `device_info` and go poorly.

There's an argument to be made that they might change these constants (so we should use an `u64` and check the mask) but given that they've added comments to the header with specific bit positions and such I don't think they would do that without changing to a `v3`.

xilinx/src/xlnx_dec_utils.rs
```diff
@@ -67,105 +67,58 @@ pub fn xlnx_calc_dec_load(xrm_ctx: xrmContext, xma_dec_props: *mut XmaDecoderPro
...
+    if let Some(device_id) = device_id {
+        device_info = (device_id << XRM_DEVICE_INFO_DEVICE_INDEX_SHIFT)
+            | ((XRM_DEVICE_INFO_CONSTRAINT_TYPE_HARDWARE_DEVICE_INDEX as u64) << XRM_DEVICE_INFO_CONSTRAINT_TYPE_SHIFT);
+    }
+
     strcpy_to_arr_i8(&mut cu_pool_prop.cuListProp.cuProps[0].kernelName, "decoder")?;
     strcpy_to_arr_i8(&mut cu_pool_prop.cuListProp.cuProps[0].kernelAlias, "DECODER_MPSOC")?;
     cu_pool_prop.cuListProp.cuProps[0].devExcl = false;
     cu_pool_prop.cuListProp.cuProps[0].requestLoad = xrm_precision_1000000_bitmask(dec_load);
+    cu_pool_prop.cuListProp.cuProps[0].deviceInfo = device_info as _;
```
scottlamb, line 84: `as _` seems unnecessary/hazardous here? If `device_info` isn't already 64-bit, I don't think the shifts above will work out well.

xilinx/src/xlnx_scal_utils.rs
```diff
@@ -112,25 +116,40 @@ pub(crate) fn xlnx_create_scal_session(
...
+        (Some(device_id), None) => {
+            let device_info = (device_id << XRM_DEVICE_INFO_DEVICE_INDEX_SHIFT)
+                | ((XRM_DEVICE_INFO_CONSTRAINT_TYPE_HARDWARE_DEVICE_INDEX as u64) << XRM_DEVICE_INFO_CONSTRAINT_TYPE_SHIFT);
+            scaler_cu_prop.deviceInfo = device_info as _;
+        }
+        (None, Some(reserve_id)) => {
+            scaler_cu_prop.poolId = reserve_id;
+        }
+        (None, None) => {
+            // use the zero indexed device as the default if no device or reserve_id have been provided
```
scottlamb, line 142: This seems like a surprising behavior. Is it necessary, or can we leave `deviceInfo` at `0`?

scottlamb (changes requested):

xilinx/src/xlnx_dec_utils.rs
```diff
@@ -15,12 +15,12 @@ pub struct XlnxDecBuffer<'a> {
...
 pub struct XlnxDecoderXrmCtx {
-    pub xrm_reserve_id: u64,
-    pub device_id: i32,
+    pub xrm_reserve_id: Option<u64>,
+    pub device_id: Option<u32>,
     pub dec_load: i32,
     pub decode_res_in_use: bool,
     pub xrm_ctx: xrmContext,
-    pub cu_list_res: xrmCuListResource,
+    pub cu_list_res: xrmCuListResourceV2,
```
scottlamb, line 23: As we discovered together yesterday: `xrmCuListPropertyV2` is enormous so needs to be boxed to avoid stack overflows. Lemme know when that's ready for review.

scottlamb (changes requested):

xilinx/src/xlnx_dec_utils.rs
epingel2: These are now all boxed.

scottlamb (approved):

PR #185, opened 2024-04-06
add v4l2 crate

ccbrown:
Inspired by Scott pointing me to the Orange Pi, this is a mostly-for-fun PR that adds a `v4l2` crate. V4L2 is the Linux kernel driver / API for capturing, encoding, and decoding video. This crate adds an implementation of our video encoder trait that uses the V4L2 API.

I developed and tested it on a [Raspberry Pi Zero 2 W](https://www.raspberrypi.com/products/raspberry-pi-zero-2-w/). The test encodes 3 seconds of 1080p30 video in about 2 seconds, i.e. it runs at 150% realtime on this device. Not bad for a $15 device.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -45,7 +45,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: test
-        args: --verbose --workspace --exclude xilinx --all-features --exclude xcoder-quadra --exclude xcoder-logan ${{ matrix.args }}
+        args: --verbose --workspace --exclude xilinx --all-features --exclude xcoder-quadra --exclude xcoder-logan --exclude v4l2 ${{ matrix.args }}
   test_xcoder_logan:
     name: Test xcoder-logan
     runs-on:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,7 +41,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: test
-        args: --verbose --workspace --exclude xilinx --all-features --exclude xcoder-quadra --exclude xcoder-logan
+        args: --verbose --workspace --exclude xilinx --all-features --exclude xcoder-quadra --exclude xcoder-logan --exclude v4l2
   test_xcoder_logan:
     name: Test xcoder-logan
     runs-on:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,6 +6,7 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-12-large]
     steps:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,6 +9,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-12-large]
+        include:
+          - os: macos-12-large
+            args: --exclude v4l2
     steps:
     - uses: actions/checkout@v1
     - name: Set Up Linux
@@ -35,14 +38,14 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: build
-        args: --verbose --workspace --exclude xilinx --features srt/async
+        args: --verbose --workspace --exclude xilinx --features srt/async ${{ matrix.args }}
     - name: Lint
-      run: cargo clippy --workspace --exclude xilinx --features srt/async --all-targets -- --deny warnings
+      run: cargo clippy --workspace --exclude xilinx --features srt/async --all-targets ${{ matrix.args }} -- --deny warnings
     - name: Test
       uses: actions-rs/cargo@v1
       with:
         command: test
-        args: --verbose --workspace --exclude xilinx --all-features --exclude xcoder-quadra --exclude xcoder-logan --exclude v4l2
+        args: --verbose --workspace --exclude xilinx --all-features --exclude xcoder-quadra --exclude xcoder-logan ${{ matrix.args }}
   test_xcoder_logan:
     name: Test xcoder-logan
     runs-on:
--- a/v4l2/Cargo.toml
+++ b/v4l2/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "v4l2"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+av-traits = { path = "../av-traits" }
+nix = { version = "0.27", features = ["fs"] }
+snafu = "0.8.0"
+v4l2r = { git = "https://github.com/Gnurou/v4l2r", rev = "6091b1f0643f20435b97413635d935052f19c739" }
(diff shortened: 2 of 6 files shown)


.github/workflows/rust.yml
```diff
@@ -6,8 +6,12 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
ccbrown, line 9: Any opposition to this? Before, if the macOS build failed in CI, the Ubuntu build would immediately be cancelled. When one build fails, I would prefer to see the other continue so I can comprehensively see where the failures are/aren't.

v4l2/src/encoder.rs
```diff
@@ -0,0 +1,401 @@
...
+        match self {
+            Self::Yuv420Planar => 3,
+            Self::Bgra => 1,
+        }
+    }
+
+    fn pixel_format(&self) -> PixelFormat {
+        PixelFormat::from_u32(match self {
+            Self::Yuv420Planar => 0x32315559,
+            Self::Bgra => 0x34524742,
```
Xaeroxe, line 85: We should create constants for these hexadecimal literals that have a meaningful and descriptive name. Additionally we should note that these constants are in fourcc format. `H264_PIXELFORMAT` a few lines below is a great example of this.

.github/workflows/rust.yml
Xaeroxe: Seems fine to me.

v4l2/src/encoder.rs
```diff
@@ -0,0 +1,401 @@
...
+            })
+            .collect::<Result<Vec<_>>>()?;
+
+        ioctl::reqbufs(&fd, QueueType::VideoCaptureMplane, MemoryType::Mmap, N_BUFFERS as _)?;
+        let capture_mappings = (0..N_BUFFERS)
+            .map(|i| -> Result<_> {
+                let buf: ioctl::QueryBuffer = ioctl::querybuf(&fd, QueueType::VideoCaptureMplane, i)?;
+                Ok(ioctl::mmap(&fd, buf.planes[0].mem_offset, buf.planes[0].length)?)
+            })
+            .collect::<Result<Vec<_>>>()?;
```
Xaeroxe, line 208: Since we have a few constant size `Vec` in here, consider turning them into arrays and using [`array::from_fn`](https://doc.rust-lang.org/std/array/fn.from_fn.html) in order to simplify usage for the CPU and the dev. 

For the `Drop` impl, you would then need to wrap these inside of a [`ManuallyDrop`](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html). This technically adds more `unsafe`, but I think that's desirable seeing as we're opinionated about the drop order anyways.

If you really don't like `ManuallyDrop`, another option is [`ArrayVec`](https://docs.rs/tinyvec/1.6.0/tinyvec/struct.ArrayVec.html).

Xaeroxe (changes requested):
Neat! Thanks for implementing this. I've added a few minor comments below.

.github/workflows/rust.yml
Xaeroxe: Seems fine to me.

v4l2/src/encoder.rs
```diff
@@ -0,0 +1,401 @@
...
+
+pub struct V4L2Encoder<F> {
+    config: V4L2EncoderConfig,
+    fd: File,
+    capture_mappings: Vec<ioctl::PlaneMapping>,
+    output_mappings: Vec<ioctl::PlaneMapping>,
+    output_format: Format,
+    pending_frames: VecDeque<F>,
+    available_output_buffers: Vec<usize>,
+    available_capture_buffers: Vec<usize>,
```
Xaeroxe, line 54: These might be a good opportunity for an [`ArrayVec`](https://docs.rs/tinyvec/latest/tinyvec/struct.ArrayVec.html)

v4l2/src/encoder.rs
```diff
@@ -0,0 +1,401 @@
...
+        let mut capture_format: Format = ioctl::g_fmt(&fd, QueueType::VideoCaptureMplane)?;
+        capture_format.pixelformat = H264_PIXELFORMAT;
+        capture_format.width = config.width as _;
+        capture_format.height = config.height as _;
+        let _: Format = ioctl::s_fmt(&mut fd, (QueueType::VideoCaptureMplane, &capture_format))?;
+
+        // set the fps if possible
+        {
+            let mut parm: bindings::v4l2_streamparm = ioctl::g_parm(&fd, QueueType::VideoOutputMplane)?;
+            let output_parm = unsafe { &mut parm.parm.output };
```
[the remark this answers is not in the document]
ccbrown: We're reading a union field here, which is always unsafe.

v4l2/src/encoder.rs
ccbrown: `tinyvec::ArrayVec` can't be used here since `PlaneMapping` doesn't implement `Default`. An array feels a little messier here as there's currently not a great way to initialize one from fallible operations ([try_map](https://doc.rust-lang.org/std/primitive.array.html#method.try_map) will help here once stable).

Instead, I went with [`arrayvec::ArrayVec`](https://docs.rs/arrayvec/0.7.4/arrayvec/struct.ArrayVec.html) both here and for `available_xxx_buffers`. Let me know how that looks to you.

v4l2/src/encoder.rs
ccbrown: `const BGRA = 0x34524742`, `Self::Bgra => BGRA` seemed a little superfluous to me. I added comments indicating that these are four-character codes and what their values are. How's that?

v4l2/src/encoder.rs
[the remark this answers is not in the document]
Xaeroxe: Thank you

v4l2/src/encoder.rs
Xaeroxe: That's a lot better. Thanks!

v4l2/src/encoder.rs
Xaeroxe: Excellent! Looks great.

v4l2/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "v4l2"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+av-traits = { path = "../av-traits" }
+nix = { version = "0.27", features = ["fs"] }
+snafu = "0.8.0"
+v4l2r = { git = "https://github.com/Gnurou/v4l2r", rev = "6091b1f0643f20435b97413635d935052f19c739" }
```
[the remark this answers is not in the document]
Xaeroxe: @ccbrown This is the only comment of mine that's outstanding now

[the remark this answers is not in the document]
ccbrown: I agree, but I'd propose we consider that beyond the scope of this PR as it applies to many (or all?) of our other third party dependencies.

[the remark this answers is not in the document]
Xaeroxe: Sure, I'll add a ticket for it to the backlog.

Xaeroxe (approved):

PR #186, opened 2024-04-07
fix core2 features for no_std build

ccbrown:
In order to build `mpeg2` with `no_std`, we need to omit default features from `core2`. Note that this was intended from the beginning – If you look at the full Cargo file, you'll see that if the `std` feature is enabled, we already enable the `core2/std` feature as well.


Xaeroxe (approved):
LGTM, maybe in the future we can add a CI check for a no_std build to avoid this.

PR #189, opened 2024-04-15
Unconditionally copy string data for netint logging

Xaeroxe:
I believe this resolves a soundness issue we're seeing with this code.

Basically we had a `Cow<'a>` where `'a` was the scope of the `rust_netint_*_callback`, and we were accidentally treating it as a `Cow<'static>`. That `Cow<'static>` was then passed through the `log` ecosystem as a format argument.


Xaeroxe:
Further investigation seems to show that this is actually fine. So I'm closing this.

PR #190, opened 2024-04-15
make rk mpp encoder re-emit parameter sets for keyframes

ccbrown:
The Rockchip encoder only emits parameter sets once at the beginning of the stream. For us to be able to join it anywhere, we need to re-emit them before all keyframes.


ccbrown:
> The code seems fine to me, though CI doesn't pass right now

Flaky SRT test. 🤔

PR #192, opened 2024-04-16
add cargo:rerun-if-changed for cc files

Xaeroxe:
Due to recent changes in `cc` this is now needed whereas it wasn't before.


scottlamb (changes requested):
I think this has the opposite effect as intended in some cases: some changes that should cause a rebuild (e.g. a change to `build.rs` itself) cause a rebuild without this change and don't with this change. If there are no `cargo:rerun-if-changed` lines, the entire package directory is implied, as described [here](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed). Adding the first one makes rebuilding *less* likely.

Can scope the change to just the two `-sys` crates of interest for now, ensure any change to the `build.rs`, the in-package `src` tree, or the logging shim all cause a rebuild?

Xaeroxe:
@scottlamb alright, I reduced the scope as requested

scottlamb (approved):

PR #193, opened 2024-04-16
update rust to 1.77.2

Xaeroxe:
This also adds a rust-toolchain.toml file, something that our dev and build environments can use to discern the correct version to compile with. If the specified toolchain is missing, rustup will automatically install it.


.github/workflows/rust.yml
```diff
@@ -28,24 +28,13 @@ jobs:
...
     - name: Build
-      uses: actions-rs/cargo@v1
-      with:
-        command: build
-        args: --verbose --workspace --exclude xilinx --features srt/async ${{ matrix.args }}
+      run: cargo build --verbose --workspace --exclude xilinx --features srt/async ${{ matrix.args }}
     - name: Lint
       run: cargo clippy --workspace --exclude xilinx --features srt/async --all-targets ${{ matrix.args }} -- --deny warnings
     - name: Test
-      uses: actions-rs/cargo@v1
```
scottlamb, line 45: This was nice:

> Warnings and errors issued by cargo will be displayed in GitHub UI

Is there a direct replacement somewhere?

x264/src/encoder.rs
```diff
@@ -244,7 +244,7 @@ mod test {
 
         assert_eq!(encoded_frames, 90);
         assert!(encoded.len() > 5000, "encoded.len() = {}", encoded.len());
-        assert!(encoded.len() < 10000, "encoded.len() = {}", encoded.len());
```
scottlamb: Was this flaky before? or did it actually change with the Rust upgrade somehow?

xcoder/xcoder-quadra/src/decoder.rs
```diff
@@ -371,51 +371,53 @@ impl<I, E> Drop for XcoderDecoder<I, E> {
 }
 
 #[cfg(test)]
-pub mod test {
```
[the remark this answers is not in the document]
Xaeroxe: `read_frames` is still under a `#[cfg(test)]`, so it won't show up in a non-test build.

Yes this was due to shadowing warnings.

x264/src/encoder.rs
Xaeroxe: Unclear. The value I was getting in CI was ~16,000. I figured a 60% improvement from a compiler upgrade was possible, so I bumped the ceiling.

.github/workflows/rust.yml
Xaeroxe: Unfortunately, no. 

So I'd be tempted to keep `actions-rs` in order to retain that. My fear with doing so is that eventually GitHub is going to bring down the hammer and remove node16 entirely. [Based on a previous announcement](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/), that could be happening very soon. So we might get forced off of `actions-rs` regardless.

x264/src/encoder.rs
scottlamb: Larger encodings are not an improvement (although they may provide higher video quality), and given that the encoding is done in hardware, shouldn't be caused by the compiler upgrade. If it is, something is really wrong with this test or the production code!

Xaeroxe: Okay, right. I had misread the test and thought `len()` was a frame count, not a byte count. The strange part is that this PR doesn't actually change anything of substance for `x264`. I'll try repeating the test without the changes from this PR.

.github/workflows/rust.yml
Xaeroxe: Actually [we have a date for that now](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/)

x264/src/encoder.rs
Xaeroxe: The commit which bumped this ceiling has been removed.

.github/workflows/rust.yml
scottlamb: I vote for one idea per PR; maybe trying `FORCE_JAVASCRIPT_ACTIONS_TO_NODE20=true` in a separate PR.

Xaeroxe: Alright, removed. I'll address that in another PR.

x264/src/encoder.rs
Xaeroxe: @scottlamb The latest test run can no longer reproduce this issue. So either it was caused by one of the commits I removed in response to https://github.com/sportsball-ai/av-rs/pull/193#discussion_r1567938110, or the issue is intermittent.

scottlamb (approved):

x264/src/encoder.rs
Xaeroxe: Figured it out https://github.com/sportsball-ai/av-rs/pull/195/commits/f1057e95c2d18ffe17e64220bfcc976fdb97f123

PR #196, opened 2024-04-17
srt fix segfault

Xaeroxe:
Fixes a segfault due to failing to cast this box correctly.

This compiled previously because the `as` pointer cast was also changing the Box type understood by the software, without actually converting the box into a fat pointer.

### DIFF
--- a/srt/src/lib.rs
+++ b/srt/src/lib.rs
@@ -591,7 +591,7 @@ where
     F: FnMut(&SrtLogEvent) + Send + Sync + 'static,
 {
     let closure_raw = Box::into_raw(Box::new(Box::new(handler) as Box<dyn FnMut(&SrtLogEvent) + Send + Sync + 'static>));
-    let closure = LogHandler(closure_raw as LogHandlerRaw);
+    let closure = LogHandler(closure_raw);
     let mut lock = LOG_HANDLER.lock().unwrap();
     let old_handler = mem::replace(&mut *lock, Some(closure));
     unsafe {
--- a/srt/src/lib.rs
+++ b/srt/src/lib.rs
@@ -938,7 +938,7 @@ mod test {
             called.fetch_add(1, Ordering::Relaxed);
         });
         let server_thread = thread::spawn(|| {
-            let listener = Listener::bind("127.0.0.1:1234").unwrap();
+            let listener = Listener::bind("127.0.0.1:1233").unwrap();
             let (mut conn, _) = listener.accept().unwrap();
             let mut buf = [0; 1316];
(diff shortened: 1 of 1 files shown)


srt/src/lib.rs
```diff
@@ -590,7 +590,7 @@ pub fn set_log_handler<F>(handler: F)
 where
     F: FnMut(&SrtLogEvent) + Send + Sync + 'static,
 {
-    let closure_raw = Box::into_raw(Box::new(Box::new(handler)));
+    let closure_raw = Box::into_raw(Box::new(Box::new(handler) as Box<dyn FnMut(&SrtLogEvent) + Send + Sync + 'static>));
     let closure = LogHandler(closure_raw as LogHandlerRaw);
```
scottlamb, line 594: Is this `as` necessary? I think it's what allowed the incorrect type to compile, so I'd prefer to kill it if possible.

scottlamb (changes requested):
Can you add a test that segfaults before and passes after? That'd give me more confidence we won't regress.

srt/src/lib.rs
Xaeroxe: Good point. Resolved in https://github.com/sportsball-ai/av-rs/pull/196/commits/578fd31447818f6f8bd239561594eea23f014b7f

Xaeroxe:
@scottlamb I added that unit test in 96f3d0a

srt/src/lib.rs
```diff
@@ -928,4 +928,31 @@ mod test {
...
+    #[test]
+    fn log_handler_used_correctly() {
+        set_log_level(LogLevel::Debug);
+        let called = Arc::new(AtomicU32::new(0));
+        let called_clone = Arc::clone(&called);
+        set_log_handler(move |_: &SrtLogEvent<'_>| {
+            called.fetch_add(1, Ordering::Relaxed);
+        });
+        let server_thread = thread::spawn(|| {
+            let listener = Listener::bind("127.0.0.1:1233").unwrap();
```
scottlamb, line 941: nit: is it possible to bind to just `:0` and then use `srt_getsockname` to get the actual port assigned? This is a pattern I use to avoid flaky tests if there are port conflicts for some reason (whether that's two `#[test]`s that both need a port and get run in parallel automatically by `cargo test`/`libtest`, someone actually running two different tests that both get some port, some long-running thing on the machine that uses the same port number, etc).

Xaeroxe: Yeah I do prefer auto ports. I addressed that in 53dd63fc8aa199d46fbb91698e145811f99df97e

scottlamb (approved):

PR #198, opened 2024-05-14
Remove outdated comments about deprecated functions

Xaeroxe:


x264/src/encoder.rs
```diff
@@ -244,7 +244,7 @@ mod test {
 
         assert_eq!(encoded_frames, 90);
         assert!(encoded.len() > 5000, "encoded.len() = {}", encoded.len());
-        assert!(encoded.len() < 10000, "encoded.len() = {}", encoded.len());
```
scottlamb, line 247: this must be based prior to 4be2fc92?

scottlamb (approved):

x264/src/encoder.rs
Xaeroxe: Ah, yeah it seems so.

PR #200, opened 2024-06-13
restructure mpegts analyzer to use much less memory and improve benchmarks

Xaeroxe:
This optimization is motivated by memory profiles we've seen come out of our dev environment. In high memory usage scenarios these vec allocations accounted for ~70% of our memory usage.

In addition to drastically reducing the memory footprint, this PR improves benchmark times. From ~1.9 ms down to ~691us

### DIFF
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -1,8 +1,9 @@
 use mpeg2::{pes, ts};
 use mpeg4::AudioDataTransportStream;
+use vecmap::VecMap;
 
 use std::{
-    collections::{HashMap, VecDeque},
+    collections::VecDeque,
     error::Error,
 };
 
@@ -413,7 +414,7 @@ enum PidState {
 
 /// Analyzer processes packets in real-time, performing cheap analysis on the streams.
 pub struct Analyzer {
-    pids: HashMap<u16, PidState>,
+    pids: VecMap<u16, PidState>,
     has_video: bool,
 }
 
@@ -421,7 +422,7 @@ impl Analyzer {
     pub fn new() -> Self {
         Self {
             pids: {
-                let mut v = HashMap::new();
+                let mut v = VecMap::new();
                 v.insert(ts::PID_PAT, PidState::Pat);
                 v
             },
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -1,7 +1,10 @@
 use mpeg2::{pes, ts};
 use mpeg4::AudioDataTransportStream;
 
-use std::{collections::VecDeque, error::Error};
+use std::{
+    collections::{HashMap, VecDeque},
+    error::Error,
+};
 
 pub type Result<T> = std::result::Result<T, Box<dyn Error + Send + Sync>>;
 
@@ -401,18 +404,16 @@ pub enum StreamInfo {
     Other,
 }
 
-#[allow(clippy::large_enum_variant)]
 #[derive(Clone)]
 enum PidState {
-    Unused,
     Pat,
     Pmt,
-    Pes { stream: Stream },
+    Pes { stream: Box<Stream> },
 }
 
 /// Analyzer processes packets in real-time, performing cheap analysis on the streams.
 pub struct Analyzer {
-    pids: Vec<PidState>,
+    pids: HashMap<u16, PidState>,
     has_video: bool,
 }
 
@@ -420,8 +421,8 @@ impl Analyzer {
     pub fn new() -> Self {
         Self {
             pids: {
-                let mut v = vec![PidState::Unused; 0x10000];
-                v[ts::PID_PAT as usize] = PidState::Pat;
+                let mut v = HashMap::new();
+                v.insert(ts::PID_PAT, PidState::Pat);
                 v
             },
             has_video: false,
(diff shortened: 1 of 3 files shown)


ccbrown:
How much of the improvement is coming from the `Box` change vs the `HashMap` change?

When I do a benchmark with only the `Box` change like...

```diff
diff --git a/mpegts-segmenter/src/analyzer.rs b/mpegts-segmenter/src/analyzer.rs
index 306e8dd..6eaad96 100644
--- a/mpegts-segmenter/src/analyzer.rs
+++ b/mpegts-segmenter/src/analyzer.rs
@@ -407,7 +407,7 @@ enum PidState {
     Unused,
     Pat,
     Pmt,
-    Pes { stream: Stream },
+    Pes { stream: Box<Stream> },
 }
 
 /// Analyzer processes packets in real-time, performing cheap analysis on the streams.
@@ -545,7 +545,7 @@ impl Analyzer {
                             if stream.is_video() {
                                 self.has_video = true;
                             }
-                            *state = PidState::Pes { stream }
+                            *state = PidState::Pes { stream: Box::new(stream) }
                         }
                     };
                 }
```

I see an even better improvement in the benchmark. 701.33 µs with this PR, 591.31 µs with only the `Box` change.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -401,27 +404,25 @@ pub enum StreamInfo {
...
     Pat,
     Pmt,
-    Pes { stream: Stream },
+    Pes { stream: Box<Stream> },
 }
 
 /// Analyzer processes packets in real-time, performing cheap analysis on the streams.
 pub struct Analyzer {
-    pids: Vec<PidState>,
+    pids: HashMap<u16, PidState>,
```
scottlamb, line 416: Did you consider another hash function? The default is kinda slow. I'm pleasantly surprised that the benchmark numbers improved with this change (I guess due to reduced CPU cache pressure? although that rarely shows up in a microbenchmark). I bet they'd improve further if we used say `fnv::FnvHashMap`. Looks like `av-rs` already uses the `fnv` crate indirectly via `h2` so it wouldn't be a totally new dep.

ccbrown:
I did a quick size check, and `PidState` is 752 bytes before the box, 16 after. So even with the `Vec`, using a box reduces the required allocations by ~98%.

mpegts-segmenter/src/analyzer.rs
```diff
@@ -401,27 +404,25 @@ pub enum StreamInfo {
...
 }
 
-#[allow(clippy::large_enum_variant)]
 #[derive(Clone)]
 enum PidState {
-    Unused,
     Pat,
     Pmt,
-    Pes { stream: Stream },
+    Pes { stream: Box<Stream> },
```
scottlamb, line 411: Just for completeness: another option would be to box the variants of `Stream`. rust-analyzer tells me `Stream::ADSAudio` is 136 bytes; `Stream::AVCVideo` is 752 bytes; `Stream::HEVCVideo` is 200 bytes; `Stream::Other` is 9 bytes. I guess in practice now we'll typically have one AVC or HEVC stream, one ADS stream, one PAT, and one PMT. So we'd save a tiny bit more by not having the audio stream sized for AVC video but not enough to go to any significant trouble for.

scottlamb:
> I did a quick size check, and `PidState` is 752 bytes before the box, 16 after. So even with the `Vec`, using a box reduces the required allocations by ~98%.

IMHO, we should use a better data structure than `vec![PidState::Unknown; 0x10000]` even with the `Box`. 0x10000 * 16 bytes = 1 MiB per incoming stream instead of 0x10000 * 752 bytes = 47 MiB is a big improvement, but I don't think we're getting much value for that 1 MiB either. There are only expected to be ~4 total (non-unknown) entries, and they can be ~consecutive to use CPU cache lines + TLB entries efficiently, as well as the improvement in total RSS.

> 701.33 µs with this PR, 591.31 µs with only the Box change.

I don't think 100 µs difference is significant in the context of the whole system. And we can speed up the hash map option by changing the hash function, which I think would be just fine. Or we could do a sorted `Vec<(pid, pidstate)>` with bisection. It's a little more trouble but should be the fastest thing in practice (in a whole program with other things putting pressure on CPU caches) given expected size.

ccbrown:
Sure. I'm not necessarily advocating for `Vec`, I'm just wondering if the `HashMap` change is actually an upgrade. Trading off a 15% performance improvement for a 1 MiB memory improvement seems like a side grade at best. If another structure or hash function can eliminate the performance tradeoff, then there's no question - we should definitely use it.

Xaeroxe:
[`vecmap-rs`](https://crates.io/crates/vecmap-rs) Seems to be the best of both worlds. On my machine it benchmarks at 498us. I've added a commit which uses that.

@scottlamb @ccbrown

mpegts-segmenter/src/analyzer.rs
Xaeroxe: Local testing didn't show any significant performance change from doing this. Plus it's a breaking change to the API. I'd be happy to reconsider that if we can show large volumes of waste coming from this though.

ccbrown (approved):
Lgtm!

mpegts-segmenter/src/analyzer.rs
Xaeroxe: I didn't see any significant performance gain from changing the hash function. The std library swapped out its implementation for [`hashbrown`](https://crates.io/crates/hashbrown) a while back, and that seems to be a big improvement over what used to be there.

scottlamb: > I didn't see any significant performance gain from changing the hash function. The std library swapped out its implementation for [hashbrown](https://crates.io/crates/hashbrown) a while back, and that seems to be a big improvement over what used to be there.

Surprising that it didn't change. The hash function and the hash table implementation are orthogonal, and the default hash function is still slow. https://doc.rust-lang.org/std/collections/struct.HashMap.html

> The default hashing algorithm is currently SipHash 1-3, though this is subject to change at any point in the future. While its performance is very competitive for medium sized keys, other hashing algorithms will outperform it for small keys such as integers as well as large keys such as long strings, though those algorithms will typically not protect against attacks such as HashDoS.

scottlamb (approved):

PR #201, opened 2024-07-11
Expose pixel format in xcoder-quadra

Xaeroxe:


xcoder/xcoder-quadra/src/lib.rs
```diff
@@ -90,6 +90,31 @@ mod linux_impl {
         let num = (fps * den as f64).round() as _;
         (num, den)
     }
+
+    #[enum_repr::EnumRepr(type = "c_uint")]
```
scottlamb, line 94: Nit: can you use `ni_pix_fmt_t` directly as the `type` and avoid the need for the `i32::try_from` in `encoder.rs`?

fyi, type aliases as enum reprs apparently will be coming natively to Rust: https://github.com/rust-lang/rfcs/pull/3659 might stick a TODO to use that when it's stable while in here.

I looked at avoiding this whole enum declaration with bindgen options, but meh. Even `newtype_enum("ni_pix_fmt_t")` is certainly not as pretty as what you have here. Although I do think when I finish up the `pixelfmt` crate, I'd like to make this use that alias (just having a fallible conversion between that to `ni_pix_fmt_t`) rather than a separate "nice" netint-specific enum. But this is probably best for now.

xcoder/xcoder-quadra/src/encoder.rs
```diff
@@ -253,6 +256,7 @@ impl<F> XcoderEncoder<F> {
                 XcoderEncoderCodec::H264 { .. } => sys::_ni_codec_format_NI_CODEC_FORMAT_H264,
                 XcoderEncoderCodec::H265 { .. } => sys::_ni_codec_format_NI_CODEC_FORMAT_H265,
             };
+            (**session).pixel_format = i32::try_from(config.pixel_format.repr()).expect("cast should never fail");
```
scottlamb, line 259: I think there are a couple things elsewhere in this function that are assuming yuv420p. Search for `ni_get_hw_yuv420p_dim`, which probably should be replaced with `ni_get_frame_dim`. Could you try adding a unit test first that uses bgra? I would expect it to fail as is.

scottlamb (changes requested):

xcoder/xcoder-quadra/src/lib.rs
Xaeroxe: > can you use ni_pix_fmt_t directly as the type and avoid the need for the i32::try_from in encoder.rs?

Unfortunately no. `ni_pix_fmt_t` is an alias for `c_uint`, which is an alias for `u32`. The `pixel_format` field is `c_int`, and that's an alias for `i32`. In C++ this ergonomic wart never really comes up but in Rust it does.

xcoder/xcoder-quadra/src/encoder.rs
Xaeroxe: Ah, dang you're right. I had deluded myself into thinking that happened downstream.

Xaeroxe:
Updating the xcoder library seems to have induced some behavior that is a stack overflow without a debugger, and a segfault with a debugger. Pending further investigation.

Xaeroxe:
Issue resolved, `_ni_xcoder_params` had grown too large to be kept on the stack.

xcoder/xcoder-quadra/src/encoder.rs
```diff
@@ -589,4 +601,57 @@ mod test {
...
+        }
+        while let Some(output) = encoder.flush().unwrap() {
+            encoded.append(&mut output.encoded_frame.expect("frame was not dropped").data);
+            encoded_frames += 1;
+        }
+
+        assert_eq!(encoded_frames, 90);
+        assert!(encoded.len() > 5000);
+
+        // To inspect the output, uncomment these lines:
```
scottlamb, line 653: Did you look at the output? I suspect it will only have correct output in the left 1/4th of the frame, due to `ni_copy_hw_yuv420p` passing a `factor` of 1 to `ni_copy_plane_data`, vs `ni_copy_frame_data` passing a `factor` of 4 for bgra.

Xaeroxe: This code still doesn't execute properly. I need to get it running under a debugger to determine why.

Xaeroxe: The doc comment for `ni_copy_frame_data` says this about `factor`

```
 *  \param[in]  factor  1 for 8 bit, 2 for 10 bit
```

So I don't think `4` is ever the correct value. Though my code does have it wrong.

scottlamb: `factor` as passed to `ni_copy_frame_data` should be 1 for 8-bit samples.

`factor` as passed in to `ni_copy_plane_data` (internally within the netint library) is 4 for rgba.

the bytes per sample for this plane. For the rgba data, it's 4 (all of red/green/blue/alpha are part of 1 sample). For the planar/semi-planar stuff, it's 1/2 for 8-bit/10-bit. (10-bit = 2 bytes because each sample is 10 bits of useful stuff and 6 bits of zeros for alignment).

```c
void ni_copy_frame_data(uint8_t *p_dst[NI_MAX_NUM_DATA_POINTERS],
                          uint8_t *p_src[NI_MAX_NUM_DATA_POINTERS],
                          int frame_width, int frame_height,
                          int factor, ni_pix_fmt_t pix_fmt,
                          int conf_win_right,
                          int dst_stride[NI_MAX_NUM_DATA_POINTERS],
                          int dst_height[NI_MAX_NUM_DATA_POINTERS],
                          int src_stride[NI_MAX_NUM_DATA_POINTERS],
                          int src_height[NI_MAX_NUM_DATA_POINTERS])
{
    // ...
    if (is_rgba)
    {
        ni_copy_plane_data(p_dst, p_src,
                             frame_width, frame_height, 4,
                             is_semiplanar, conf_win_right,
                             dst_stride, dst_height,
                             src_stride, src_height,
                             0); // just one plane for rgba
    }
    else
    {
        ni_copy_hw_yuv420p(p_dst, p_src, frame_width, frame_height, factor,
                           is_semiplanar, conf_win_right, dst_stride,
                           dst_height, src_stride, src_height);
    }
}
```

Xaeroxe: I have now looked at the output. The current version is correct.

Xaeroxe:
Tests pass now, the bgra output looks good upon being inspected.

scottlamb (approved):

PR #210, opened 2024-11-20
Revert "Use new pts type for decoder in accordance with new xrt"

Xaeroxe:
Reverts sportsball-ai/av-rs#209

I guess we can't use 2.11.730 anymore.


Xaeroxe:
Disregard

PR #211, opened 2024-11-27
release object more explicitly

scottlamb:
...to avoid build failure due to the following lint error, I assume since some compiler upgrade.

```
warning: foundation@0.1.0: src/lib.mm:9:12: warning: variable 'm' set but not used [-Wunused-but-set-variable]
warning: foundation@0.1.0:         id m = (__bridge_transfer id)obj;
warning: foundation@0.1.0:            ^
warning: foundation@0.1.0: 1 warning generated.
```

My Objective-C is quite rusty (pardon the pun), but my understanding is...

* `__bridge_transfer` converts a `CFObject` to a `NSObject`.
* `@autoreleasepool` releases the object on leaving the block.

Let's just call `CFRelease` instead.


Xaeroxe (approved):
LGTM!

PR #214, opened 2025-01-14
Make API safer and simpler

Xaeroxe:
This PR has several breaking changes that are all related to removing footguns and making the API easier to use correctly.

- Stop using `simple_error` and swap out for a `thiserror` based implementation. This makes it easier to process these errors programmatically.
- Remove `Xlnx*XrmCtx` types from the public API. These types have a few usability warts.
  - Their lifetime is necessarily tied to the corresponding `XlnxDecoder`, `XlnxEncoder`, or `XlnxScaler`. So you can't drop the `Xlnx*XrmCtx` without first having dropped the session type.
  - Each `Xlnx*XrmCtx` can only support one downstream session at a time. As-is using the context to create an additional session will actually leak compute unit reservations.
  
  So it makes more sense to remove them from the public API, and then make them part of the downstream session object. This makes it easier to guarantee that the `Xlnx*XrmCtx` is interacted with in a correct manner.
  
- Remove functions that allocate compute units from the public API, instead call these when the relevant session object is created. This made more sense to me as the corresponding release of compute units happens when the `Xlnx*XrmCtx` object is released anyways.

- Move some public properties to private, add getters for them. This closes a gap where it was possible to modify these properties after they'd been set correctly or sent to Xilinx APIs.

Ideas for future work

- There's a lot of repetitive code here. Can we use generics and traits more to reduce repetition?

### DIFF
--- a/xilinx/src/xlnx_dec_utils.rs
+++ b/xilinx/src/xlnx_dec_utils.rs
@@ -1,9 +1,8 @@
 use std::{ffi::CString, os::raw::c_char, str::from_utf8};
 
 use libloading::{Library, Symbol};
-use simple_error::{bail, SimpleError};
 
-use crate::{strcpy_to_arr_i8, sys::*, xrm_precision_1000000_bitmask, XrmContext};
+use crate::{strcpy_to_arr_i8, sys::*, xrm_precision_1000000_bitmask, Error, XlnxError, XrmContext, XrmPlugin};
 
 const DEC_PLUGIN_NAME: &[u8] = b"xrmU30DecPlugin\0";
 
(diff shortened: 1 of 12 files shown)


Xaeroxe:
Ah, worth noting that `XlnxTranscodeXrmCtx` is an exception to everything the prior comment said about `Xlnx*XrmCtx` types.

xilinx/src/xlnx_decoder.rs
```diff
@@ -1,27 +1,38 @@
...
-use simple_error::{bail, SimpleError};
 
 pub struct XlnxDecoder<'a> {
     pub dec_session: *mut XmaDecoderSession,
     pub frame_props: XmaFrameProperties,
     in_buf: XmaDataBuffer,
     pub out_frame: *mut XmaFrame,
     pub flush_sent: bool,
-    decoder_ctx_lifetime: PhantomData<XlnxDecoderXrmCtx<'a>>,
+    xlnx_dec_ctx: ManuallyDrop<XlnxDecoderXrmCtx<'a>>,
```
epingel2, line 15: Manually Drop is an excellent option for this!

xilinx/src/xlnx_decoder.rs
```diff
@@ -1,27 +1,38 @@
...
 
 impl<'a> XlnxDecoder<'a> {
-    pub fn new(xma_dec_props: &mut XmaDecoderProperties, xlnx_dec_ctx: &mut XlnxDecoderXrmCtx<'a>) -> Result<Self, SimpleError> {
-        let dec_session = xlnx_create_dec_session(xma_dec_props, xlnx_dec_ctx)?;
+    pub fn new(
+        xrm_ctx: &'a XrmContext,
+        xma_dec_props: &mut XmaDecoderProperties,
+        device_id: Option<u32>,
+        reserve_id: Option<u64>,
+        dec_load: i32,
```
epingel2, line 24: The load is required here to make a decoder.

xilinx/src/xlnx_dec_utils.rs
```diff
@@ -38,30 +37,21 @@ impl<'a> XlnxDecoderXrmCtx<'a> {
 
 /// Calculates the decoder load uing the xrmU30Dec plugin.
 #[allow(clippy::not_unsafe_ptr_arg_deref)]
-pub fn xlnx_calc_dec_load(xrm_ctx: &XrmContext, xma_dec_props: *mut XmaDecoderProperties) -> Result<i32, SimpleError> {
+pub(crate) fn xlnx_calc_dec_load(xrm_ctx: &XrmContext, xma_dec_props: *mut XmaDecoderProperties) -> Result<i32, Error> {
```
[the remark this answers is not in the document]
epingel2: Oh did you mean to make the context `pub(crate)` and not this function?

xilinx/src/xlnx_error.rs
```diff
@@ -58,3 +58,61 @@ impl fmt::Debug for XlnxError {
         write!(f, "XlnxError {{ err: {:?}, message: {:?} }}", self.err, self.message,)
     }
 }
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
epingel2, line 63: This Looks Excellent! good work

xilinx/src/xlnx_scaler.rs
```diff
@@ -1,23 +1,30 @@
...
-use simple_error::SimpleError;
+use crate::{sys::*, xlnx_error::*, xlnx_scal_utils::*, XrmContext};
 
 pub const SCAL_MAX_ABR_CHANNELS: usize = 8;
 pub const SCAL_MAX_SESSIONS: usize = 2;
 pub const SCAL_RATE_STRING_LEN: usize = 8;
 
 pub struct XlnxScaler<'a> {
     scal_session: *mut XmaScalerSession,
-    pub out_frame_list: Vec<*mut XmaFrame>,
```
epingel2, line 12: I don't see a new accessor for the list of output frames. Or is one to expect to pass a list in which is mutated and then read? It's not entirely clear

epingel2 (changes requested):
Just a few things here. Over all This looks like a big improvement

xilinx/src/xlnx_dec_utils.rs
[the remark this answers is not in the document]
Xaeroxe: Yeah that was a mistake clearly. I'll fix it

Xaeroxe:
@epingel2 addressed in https://github.com/sportsball-ai/av-rs/pull/214/commits/6e2bfe9c590227a22adbefe9c0bb3b8f2f187484

epingel2 (approved):
Looks good!

PR #215, opened 2025-01-21
More xilinx improvements

Xaeroxe:
- Add APIs to the `XrmContext` which will tell how many CUs are available for each kernel type.
- Remove unused `XlnxDecBuffer` structure.

### DIFF
--- a/xilinx/src/xrm_context.rs
+++ b/xilinx/src/xrm_context.rs
@@ -1,9 +1,12 @@
-use crate::{sys, xlnx_fill_dec_pool_props, xlnx_fill_enc_pool_props, xlnx_fill_scal_cu_props, xrmCheckCuAvailableNumV2, xrmCheckCuPoolAvailableNumV2, xrmCuPoolPropertyV2, xrmCuPropertyV2, Error};
+use crate::{sys, xlnx_fill_dec_pool_props, xlnx_fill_enc_pool_props, xlnx_fill_scal_pool_props, xrmCheckCuPoolAvailableNumV2, xrmCuPoolPropertyV2, Error};
 
 /// A wrapper of the XrmContext which will correctly drop the context when dropped
 pub struct XrmContext {
     context: sys::xrmContext,
 }
+
+unsafe impl Send for XrmContext {}
+// Sync is omitted on purpose as I have no reason to believe these APIs are thread safe.
 
 impl XrmContext {
     pub fn new() -> Self {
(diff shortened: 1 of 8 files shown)


xilinx/src/xrm_context.rs
```diff
@@ -15,6 +18,28 @@ impl XrmContext {
     pub unsafe fn raw(&self) -> sys::xrmContext {
         self.context
     }
+
+    pub fn scal_cu_available(&self) -> Result<i32, Error> {
```
scottlamb, line 22: I rather hate Xilinx's docs here; I think we should have our own docstrings. Am I correctly understanding: this returns the number of CUs available in the default pool that have 1 load unit available?

Is that the operation we want? will we always be querying for 1?

same question for the other calls.

Xaeroxe: So a couple things to note here, and you're right this should be documented better so I'll author some of this for doc comments.

1. This is exclusively for monitoring, not making reservations. Reservations take a different but similar code path.
2. Pools are somewhat confusingly named. A pool is more like a pre-reservation for slots. There is no default pool.

So what I'm doing here is asking "If I needed to allocate a pool with these parameters and the smallest possible load, how many of those pools would I be able to allocate right now?"

I use the smallest possible load to ensure our monitoring gives us the most fine-grained result.

For scaling we're only requesting a single kernel type, so asking for it in the form of a pool is dubiously required. However the other two work loads require two kernel types, so asking for it in the form of a pool is the only way to get a good answer.

In practice, on an unloaded system, this comes back with 2000 scaling CUs, 64 decoding pools, and 64 encoding pools.

scottlamb: Looking forward to the doc comments. I think we will just have to develop our own documentation to have any hope of a team understanding of how this stuff works given the Xilinx docs' poor quality.

> There is no default pool.

I was getting the idea of a default pool from the "the system default pool id is 0" all over [https://xilinx.github.io/XRM/lib.html]. What does that mean if there's no default pool?

xrmCheckCuPoolAvailableNum also says "given a pool of kernels’s property with kernel name or alias or both and request load." but on second read I think that just means that the supplied `xrmCuPoolProperty*` argument is used to specify those properties, not that it's given a pool or anything. Their grammar is not aiding my understanding. :-(

> For scaling we're only requesting a single kernel type, so asking for it in the form of a pool is dubiously required. However the other two work loads require two kernel types, so asking for it in the form of a pool is the only way to get a good answer.

Makes sense to me to treat all three types in the same way, even if scaling could be handled in a simpler way.

Xaeroxe: Okay, so yeah I had to rethink my approach too. It seems the default pool is a special pool. The default pool contains all capacity that is not reserved for a different pool.

Xaeroxe: Added more documentation in 3e4434b5ce479a60e528a28ce9535eca852db678

scottlamb (approved):

PR #223, opened 2025-10-17
Add support to create decompression sessions with destination image buffer attributes ex. specify pixel format

epingel2:
Allows decompression sessions to be created with a "destination_image_buffer_attributes" dictionary.
This allows the caller to specify useful things such as output dimensions and output pixel format. 

I have implemented this with a separate `new` function added to the decompression session so as to preserve backwards compatibility, but if that were not an issue, this could also be added as an option to the existing `new` function.

### DIFF
--- a/macos/video-toolbox/build.rs
+++ b/macos/video-toolbox/build.rs
@@ -18,6 +18,7 @@ fn main() {
             .allowlist_function("VTSessionCopySupportedPropertyDictionary")
             .allowlist_function("VTSessionSetProperty")
             .allowlist_var("kCMTime.+")
+            .allowlist_var("kCVPixelBufferPixelFormatTypeKey")
             .allowlist_var("kCVPixelFormatType_.+")
             .allowlist_var("kCMSampleAttachmentKey_.+")
             .allowlist_var("kVTVideoEncoderList_.+")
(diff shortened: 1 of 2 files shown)


ccbrown:
Hey, thanks for the PR! This looks good, and I kind of like keeping two constructors, but I have one suggestion: Since `new` and `new_with_destination_image_buffer_attributes` are kind of long but almost identical, maybe it would make sense to deduplicate that code by creating a private function that they both call:

```rs
fn new_impl(
    format_desc: &VideoFormatDescription,
    destination_image_buffer_attributes: Option<Dictionary>,
) -> Result<Self, OSStatus> {
```

epingel2:
> Hey, thanks for the PR! This looks good, and I kind of like keeping two constructors, but I have one suggestion: Since `new` and `new_with_destination_image_buffer_attributes` are kind of long but almost identical, maybe it would make sense to deduplicate that code by creating a private function that they both call:
> 
> ```rust
> fn new_impl(
>     format_desc: &VideoFormatDescription,
>     destination_image_buffer_attributes: Option<Dictionary>,
> ) -> Result<Self, OSStatus> {
> ```

Yes. excellent advice. Done!

ccbrown (approved):

